PR #9, opened 2022-03-29
Init openbsd / netbsd image creation

### LINKED ISSUE #6: openbsd: Build Image

nbyavuz:
Fixes #6 and #7.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,6 +12,7 @@ env:
   # the following variables are chosen in .cirrus.star, based on the branch
   # being built etc
   # BRANCH_TYPE = dev|main
+  # BUCKET = ${GCP_PROJECT_MAIN}-bucket or ${GCP_PROJECT_DEV}-bucket
   # GCP_PROJECT = $GCP_PROJECT_MAIN or $GCP_PROJECT_DEV
   # GCP_REGION
   # GCP_REPO
@@ -123,6 +124,60 @@ task:
       pg-ci-${CIRRUS_TASK_NAME}-${DATE}
 
 
+task:
+  matrix:
+    - name: openbsd-7-1-vanilla
+      env:
+        PACKERFILE: packer/bsd/openbsd_vanilla.pkr.hcl
+
+    - name: openbsd-7-1-postgres
+      depends_on:
+        - openbsd-7-1-vanilla
+      env:
+        PACKERFILE: packer/bsd/openbsd_postgres.pkr.hcl
+
+    - name: netbsd-9-2-vanilla
+      env:
+        PACKERFILE: packer/bsd/netbsd_vanilla.pkr.hcl
+
+    - name: netbsd-9-2-postgres
+      depends_on:
+        - netbsd-9-2-vanilla
+      env:
+        PACKERFILE: packer/bsd/netbsd_postgres.pkr.hcl
+
+  container:
+    dockerfile: docker/linux_debian_packer
+    cpu: 2
+    memory: 4G
+
+  skip: $CIRRUS_LAST_GREEN_CHANGE != '' && $CIRRUS_CRON != 'regular-rebuild' && !changesInclude('.cirrus.yml', 'docker/linux_debian_packer', 'files/bsd/*', 'scripts/bsd/*', $PACKERFILE)
+  auto_cancellation: false
+
+  <<: *gcp_auth_unix
+
+  info_script:
+    - packer --version
+
+  build_image_script: |
+    DATE=$(date --utc +'%Y-%m-%dt%H-%M')
+    PACKER_LOG=1
+
+    packer build \
+      -timestamp-ui \
+      -force \
+      -var "image_date=$DATE" \
+      -var "image_name=${CIRRUS_TASK_NAME}" \
+      -var "bucket=$BUCKET" \
+      -var gcp_project=$GCP_PROJECT \
+      "$PACKERFILE"
+    gcloud compute images add-iam-policy-binding \
+      --project "$GCP_PROJECT" \
+      --member=allAuthenticatedUsers \
+      --role=roles/compute.imageUser \
+      pg-ci-${CIRRUS_TASK_NAME}-${DATE}
+
+
 # Clean up old images regularly - after a while the individually small cost
 # starts adding up
 #
@@ -141,6 +196,8 @@ task:
 
   depends_on:
     - freebsd-13-0
+    - netbsd-9-2-postgres
+    - openbsd-7-1-postgres
     - bullseye
     - sid
     - sid-newkernel
--- a/docker/linux_debian_packer
+++ b/docker/linux_debian_packer
@@ -17,5 +17,8 @@ RUN \
   apt-get update -y && \
   apt-get install --no-install-recommends -y \
     google-cloud-sdk \
+    \
+    qemu-system \
+    qemu-utils \
     && \
   apt-get clean && rm -rf /var/lib/apt/lists/*
--- a/files/bsd/rc.local.sh
+++ b/files/bsd/rc.local.sh
@@ -1,0 +1,145 @@
+#!/bin/sh
+
+# Save to /etc/rc.local, to be ran on startup
+
+if [ "$(uname)" = "NetBSD" ]
+then
+    export PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/pkg/bin:/usr/pkg/sbin
+
+    # restarting syslogd
+    restart_syslogd()
+    {
+        /etc/rc.d/syslogd restart > /dev/null
+    }
+    
+    # chown home paths
+    chown_home()
+    {
+        chown -R ${username}:users /home/${username}
+    }
+elif [ "$(uname)" = "OpenBSD" ]
+then
+    export PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/local/sbin
+
+    # restarting syslogd
+    restart_syslogd()
+    {
+        rcctl restart syslogd > /dev/null
+    }
+
+    # chown home paths
+    chown_home()
+    {
+        chown -R ${username}:${username} /home/${username}
+    }
+else
+    # Unsupported OS
+    echo "Unsupported OS, exiting"
+    exit 1
+fi
+
+### get instance hostname ###
+
+instance_hostname=$(curl -s -H "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/instance/name)
+current_hostname=$(hostname)
+
+if [ "$instance_hostname" != "" ]
+then
+    if [ "$instance_hostname" != "$current_hostname" ]
+    then
+        echo "Setting hostname to $instance_hostname"
+        hostname $instance_hostname
+        echo $instance_hostname > /etc/myname
+        restart_syslogd
+    else
+        echo "Hostname is correct"
+    fi
+else
+    echo "Could not discover hostname"
+fi
+
+### set up user's keys ###
+
+# There are two types of ssh keys on the GCE; project level and instance level ssh keys.
+# Fetch both of the keys and put them to the related paths
+project_level_ssh_keys=$(curl -s -H "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/project/attributes/ssh-keys)
+instance_level_ssh_keys=$(curl -s -H "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/instance/attributes/ssh-keys)
+ssh_keys=$(printf '%s\n%s' "$project_level_ssh_keys" "$instance_level_ssh_keys")
+
+if [ "$ssh_keys" != "" ]
+then
+    echo "$ssh_keys" | while read line
+    do
+        username="$(echo $line | cut -d: -f1)"
+        user_key="$(echo $line | cut -d: -f2-)"
+        key_comment="$(echo $line | awk '{print $NF}')"
+
+        if [ "$username" = "root" ]
+        then
+            mkdir -p /root/.ssh
+            touch /root/.ssh/authorized_keys
+            if [ $(grep -c "$user_key" /root/.ssh/authorized_keys) -eq 0 ]
+            then
+                echo "$user_key" >> /root/.ssh/authorized_keys
+                chmod 600 /root/.ssh/authorized_keys
+                echo "$username: added ssh-key $key_comment"
+            else
+                echo "$username: ssh-key $key_comment already exists"
+            fi
+        elif [ "$username" != "" ]
+        then
+            useradd ${username}
+            mkdir -p /home/${username}/.ssh
+            touch /home/${username}/.ssh/authorized_keys
+            chown_home
+            if [ $(grep -c "$user_key" /home/${username}/.ssh/authorized_keys) -eq 0 ]
+            then
+                echo "$user_key" >> /home/${username}/.ssh/authorized_keys
+                chmod 600 /home/${username}/.ssh/authorized_keys
+                echo "$username: added ssh-key $key_comment"
+            else
+                echo "$username: ssh-key $key_comment already exists"
+            fi
+        fi
+    done
+    # restart sshd
+    /etc/rc.d/sshd restart
+else
+    echo "No keys found"
+fi
+
+### Check for bootstrap scripts ###
+
+instance_bootstrap=$(curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/attributes/bootstrap)
+
+if [ "$instance_bootstrap" != "" ]
+then
+    if [ $(echo "$instance_bootstrap" | grep -c 'was not found') -eq 0 ]
+    then
+        if [ ! -f /var/log/bootstrap ] || [ $(grep -c "$instance_bootstrap" /var/log/bootstrap) -eq 0 ]
+        then
+            echo "Bootstrap: starting"
+            ftp -V -o - $instance_bootstrap | sh
+            echo $instance_bootstrap >> /var/log/bootstrap
+        else
+            echo "Bootstrap: completed"
+        fi
+    else
+        echo "Bootstrap: metadata not found"
+    fi
+else
+    echo "Bootstrap: empty curl"
+fi
+
+### Run startup scripts ###
+
+startup_script=$(curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/attributes/startup-script)
+
+if [ "$startup_script" != "" ]
+then
+    if [ $(echo "$startup_script" | grep -c 'was not found') -eq 0 ]
+    then
+        echo "Running startup script"
+        echo "${startup_script}" | sh > /dev/null 2>&1 &
+    fi
+fi
--- a/gcp_project_setup.txt
+++ b/gcp_project_setup.txt
@@ -26,6 +26,9 @@ gcloud iam service-accounts keys create \
 # FIXME: It'd be better to create a more restrictive role with just the
 # permissions we need, intead of using the pre-defined roles.
 
+# Create bucket for uploading images of openBSD and netBSD, bucket name
+# should be unique
+gsutil mb -p $GCP_PROJECT -b on gs://$GCP_PROJECT-bucket
 
 # For packer we need:
 # https://www.packer.io/docs/builders/googlecompute
--- a/packer/bsd/netbsd_vanilla.pkr.hcl
+++ b/packer/bsd/netbsd_vanilla.pkr.hcl
@@ -1,0 +1,127 @@
+variable "bucket" { type = string }
+variable "gcp_project" { type = string }
+variable "image_date" { type = string }
+variable "image_name" { type = string }
+
+variable "prefix" {
+  type = string
+  default = ""
+}
+
+source "qemu" "qemu-gce-builder" {
+
+  boot_command            = [
+    # Installation messages in English
+    "<enter><wait5s>",
+    # Keyboard type: unchanged
+    "<enter><wait5s>",
+    # Install NetBSD to hard disk
+    "<enter><wait5s>",
+    # Continue: Yes
+    "<down><enter><wait5s>",
+    # Available disk: Select default
+    "<enter><wait5s>",
+    # Guid Partition Table (GPT)
+    "<enter><wait5s>",
+    # This is the correct geometry
+    "<enter><wait5s>",
+    # Use default partition sizes
+    "<down><enter><wait5s>",
+    # Partition sizes ok
+    "<enter><wait5s>",
+    # Continue: Yes
+    "<down><enter><wait5s>",
+    # Use serial port com0 and continue
+    "b<enter><wait5s>x<enter><wait5s>",
+    # Installation without X11
+    "<down><enter><wait5s>",
+    # Install from CD-ROM / DVD / install image media
+    "<enter><wait300s>",
+    # Continue
+    "<enter><wait5s>",
+    # Enable sshd
+    "g<enter><wait5s>",
+    # Change root password
+    "d<enter><wait5s><enter><wait5s>",
+    # Set root password
+    "packer<enter><wait5s>",
+    # Set root password again
+    "packer<enter><wait5s>",
+    # Set root password again
+    "packer<enter><wait5s>",
+    # Finished configuring
+    "x<enter><wait5s>",
+    # Continue
+    "<enter><wait5s>",
+    # Reboot the computer
+    "d<enter><wait180s>",
+    # Login as a root
+    "root<enter><wait5s>packer<enter><wait5s>",
+    # Allow root username/password login
+    "sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/g' /etc/ssh/sshd_config<enter><wait5s>",
+    # Enable ipv4 network(ipv6 has problems with QEMU)
+    "echo '/sbin/dhcpcd -4' > /etc/rc.local<enter><wait5s>",
+    "reboot<enter>"
+  ]
+
+  boot_wait               = "120s"
+  cpus                    = 2
+  disk_size               = 25600
+  memory                  = 1024
+  headless                = true
+  iso_checksum            = "sha256:5f1bca14c4090122f31713dd86a926f63109dd6fb3c05f9b9b150a78acc8bc7e"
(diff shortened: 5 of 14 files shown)


scripts/bsd/netbsd-prep-postgres.sh
```diff
@@ -0,0 +1,27 @@
...
+bison \
+ccache \
+p5-IPC-Run \
+flex \
+pkgconf \
+python38 \
+icu \
+lz4 \
+libxslt \
+tcl && \
```
anarazel, line 26: I think it might be a good idea to have two different "levels" of images. One that just defines a basic working open/netbsd GCP image, and then a second one installing all the postgres dependencies, ontop of the prior image. That way other projects can use the base image (including hopefully getting the cirrus folks to automatically test their support for open / netbsd).

files/bsd/openbsd-rc.local.sh
```diff
@@ -0,0 +1,85 @@
...
+            echo "Bootstrap: completed"
+        fi
+    else
+        echo "Bootstrap: metadata not found"
+    fi
+else
+    echo "Bootstrap: empty curl"
+fi
+
+/sbin/sysctl kern.seminfo.semmni=2048 && \
```
anarazel, line 84: I think this should be done separately - it's really unrelated to the gcp integration.

packer/netbsd.pkr.hcl
```diff
@@ -0,0 +1,137 @@
...
+  post-processors {
+
+    post-processor "manifest" {
+      output     = "output/manifest.json"
+      strip_path = true
+    }
+
+    post-processor "shell-local" {
+      inline = [
+        "jq -r '.builds[] | (.name + \"/\" + .files[].name)' output/manifest.json | grep '\\.vmdk$' > output/disk-path", "if [ $(wc -l < output/disk-path) -gt 1 ]; then exit 1; fi", "qemu-img convert -f vmdk -O raw \"output-$(cat output/disk-path)\" output/disk.raw", "rm output/manifest.json output/disk-path"
```
anarazel, line 117: That's very hard to read. Perhaps you could format it more like the existing packer scripts? E.g.
```
    inline = [
      <<-SCRIPT
        cat /boot/loader.conf || true
        PAGER=cat freebsd-update fetch install
...
```

packer/netbsd.pkr.hcl
```diff
@@ -0,0 +1,137 @@
...
+      ]
+    }
+
+    post-processor "artifice" {
+      files = ["output/disk.raw"]
+    }
+
+    post-processor "compress" {
+      output = "output/netbsd92.tar.gz"
```
anarazel, line 126: How much of an effect does changing the compression level here have?

packer/netbsd.pkr.hcl
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    post-processor "googlecompute-import" {
+      gcs_object_name   = "packer-${var.image_name}-${var.image_date}.tar.gz"
+      bucket            = "${var.bucket}"
+      image_family      = "pg-ci-netbsd-9-2"
```
anarazel, line 132: Can we define the name a bit more centrally in a local variable?

packer/netbsd.pkr.hcl
```diff
@@ -0,0 +1,137 @@
...
+      "packer<enter><wait10s>",
+      "x<wait2s><enter><wait10s>",
+      "<enter><wait10s>",
+      "d<wait2s><enter><wait120s>",
+      "r<wait2s>o<wait2s>o<wait2s>t<wait2s>",
+      "<enter><wait10s>",
+      "p<wait2s>a<wait2s>c<wait2s>k<wait2s>e<wait2s>r<wait2s>",
+      "<enter><wait10s>",
+      "sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/g' /etc/ssh/sshd_config<enter><wait10s>",
+      "sed -i 's/#PasswordAuthentication yes/PasswordAuthentication yes/g' /etc/ssh/sshd_config<enter><wait10s>",
```
anarazel, line 50: Why do we need password based login?

.cirrus.yml
```diff
@@ -123,6 +126,64 @@ task:
...
+    - name: netbsd-9-2
+      env:
+        PACKERFILE: packer/netbsd.pkr.hcl
+
+  compute_engine_instance:
+    image_project: pg-ci-images-dev
+    image: debian-bullseye-nv
+    platform: linux
+    cpu: 4
+    memory: 16G
```
anarazel, line 144: Do we really need this much memory? That'll reduce the number of concurrent cirrus tasks we do noticeably.

.cirrus.yml
```diff
@@ -123,6 +126,64 @@ task:
...
+    disk: 25
+    nested_virtualization: true
+    use_ssd: true
```
anarazel, line 147: This isn't the default?

.cirrus.yml
```diff
@@ -123,6 +126,64 @@ task:
...
+  auto_cancellation: false
+
+  <<: *gcp_auth_unix
+
+  install_kernel_script: |
+    apt-get -y update
+    apt-get -y upgrade
+    apt-get -y dist-upgrade
+    apt-get install -y build-essential
+    apt-get install -y linux-headers-$(uname -r)
```
anarazel, line 160: I think this should do `export DEBIAN_FRONTEND=noninteractive` before, to get rid of some of the noisy output.

packer/netbsd.pkr.hcl
```diff
@@ -0,0 +1,137 @@
...
+      "sed -i 's/UsePam yes/UsePam no/g' /etc/ssh/sshd_config<enter><wait10s>",
+      "echo '/sbin/dhcpcd' > /etc/rc.local<enter><wait10s>",
+      "reboot<enter>"
+    ]
+
+  boot_wait               = "180s"
+  cpus                    = 4
+  disk_size               = 25600
+  memory                  = 4096
```
anarazel, line 59: Do we need 4GB / does it makes things faster?

files/bsd/openbsd-rc.local.sh
```diff
@@ -0,0 +1,82 @@
...
+    then
+        echo "Setting hostname to $instance_hostname"
+        hostname $instance_hostname
+        echo $instance_hostname > /etc/myname
+        rcctl restart syslogd > /dev/null
+    else
+        echo "Hostname is correct"
+    fi
+else
+        echo "Could not discover hostname"
```
anarazel, line 22: Indentation looks wrong.

files/bsd/openbsd-rc.local.sh
```diff
@@ -0,0 +1,82 @@
...
+            /usr/sbin/useradd ${username}
+            mkdir -p /home/${username}/.ssh
+            touch /home/${username}/.ssh/authorized_keys
+            chown -R ${username}:${username} /home/${username}
+
+            if [ "$(grep -c "$user_key" /home/${username}/.ssh/authorized_keys)" == "0" ]
+            then
+                echo "$user_key" >> /home/${username}/.ssh/authorized_keys
+                # Add ssh key to root
+                echo "$user_key" >> /root/.ssh/authorized_keys
```
anarazel, line 48: Why are the keys added both to the user and to root?

files/bsd/openbsd-rc.shutdown.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# fetch and run shutdown scripts
+(/usr/local/bin/curl -s -L "http://metadata.google.internal/computeMetadata/v1/instance/attributes/shutdown-script" -H "Metadata-Flavor: Google") | /bin/sh
```
anarazel, line 4: What's the point of the subshell around curl?

.cirrus.yml
```diff
@@ -141,6 +229,8 @@ task:
 
   depends_on:
     - freebsd-13-0
+    - netbsd-9-2-postgres
+    - openbsd-7-0-postgres
```
anarazel, line 233: 7.1 was just released...

docker/linux_debian_packer
```diff
@@ -17,5 +17,8 @@ RUN \
   apt-get update -y && \
   apt-get install --no-install-recommends -y \
     google-cloud-sdk \
+    \
+    qemu-system \
+    qemu-utils \
```
anarazel, line 22: Nice, thanks for moving to qemu from virtualbox...

files/bsd/netbsd-rc.local.sh
```diff
@@ -0,0 +1,83 @@
...
+    then
+        echo "Setting hostname to $instance_hostname"
+        hostname $instance_hostname
+        echo $instance_hostname > /etc/myname
+        /etc/rc.d/syslogd restart > /dev/null
+    else
+        echo "Hostname is correct"
+    fi
+else
+        echo "Could not discover hostname"
```
anarazel, line 22: Misleading indentation.

files/bsd/netbsd-rc.local.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/sh
+
+# Save to /etc/rc.local, to be ran on startup
```
anarazel, line 3: The scripts for net and openbsd are quite similar. What would it take to share them?

files/bsd/openbsd-rc.local.sh
```diff
@@ -0,0 +1,82 @@
...
+            else
+                echo "$username: ssh-key $key_comment already exists"
+            fi
+        fi
+    done
+else
+    echo "No keys found"
+fi
+
+### Check for bootstrap value ###
```
anarazel, line 60: s/value/scripts/?

files/bsd/openbsd-rc.local.sh
```diff
@@ -0,0 +1,82 @@
...
+
+if [ "$instance_bootstrap" != "" ]
+then
+    if [ "$(echo "$instance_bootstrap" | grep -c 'was not found')" != "1" ]
+    then
+        if [ ! -f /var/log/bootstrap ] || [ "$(grep -c "$instance_bootstrap" /var/log/bootstrap)" == "0" ]
+        then
+            echo "Bootstrap: starting"
+            export PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/local/sbin
+            ftp -V -o - $instance_bootstrap | sh
```
anarazel, line 72: Yikes, this stuff is shared via ftp???

gcp_project_setup.txt
```diff
@@ -26,6 +26,8 @@ gcloud iam service-accounts keys create \
 # FIXME: It'd be better to create a more restrictive role with just the
 # permissions we need, intead of using the pre-defined roles.
 
+# Create pg-ci-bucket for uploading images of openBSD and netBS
```
anarazel, line 29: *netBSD

scripts/bsd/netbsd-set-cron.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# Set cron job to get and run startup scripts
+crontab -r
+(crontab -l 2>/dev/null; echo '@reboot (/usr/pkg/bin/curl -s -L "http://metadata.google.internal/computeMetadata/v1/instance/attributes/startup-script" -H "Metadata-Flavor: Google") | /bin/sh') | crontab -
```
anarazel, line 5: Why is this done via cron? There should at least be a comment explaining this...

packer/bsd/openbsd_vanilla.pkr.hcl
```diff
@@ -0,0 +1,88 @@
...
+    source = "files/bsd/openbsd-rc.shutdown.sh"
+    destination = "/etc/rc.shutdown"
+  }
+
+  provisioner "shell" {
+    inline = ["chmod 744 /etc/rc.local && chmod 744 /etc/rc.shutdown"]
+  }
+
+  provisioner "shell" {
+    script = "scripts/bsd/openbsd-set-cron.sh"
```
anarazel, line 72: FWIW, stuff like this can just be done inline here and be at least as readable...

.cirrus.yml
```diff
@@ -123,6 +126,91 @@ task:
       pg-ci-${CIRRUS_TASK_NAME}-${DATE}
 
 
+task:
+  matrix:
+    - name: openbsd-7-0-vanilla
+      env:
+        PACKERFILE: packer/bsd/openbsd_vanilla.pkr.hcl
+
+    - name: netbsd-9-2-vanilla
```
anarazel, line 135: Could these be part of the existing packer invocation? Then we'd not have to repeat all the boilerplate again?

.cirrus.yml
```diff
@@ -123,6 +126,91 @@ task:
...
+      --project "$GCP_PROJECT" \
+      --member=allAuthenticatedUsers \
+      --role=roles/compute.imageUser \
+      pg-ci-${CIRRUS_TASK_NAME}-${DATE}
+
+
+task:
+  matrix:
+    - name: openbsd-7-0-postgres
+      depends_on:
```
anarazel, line 174: I think these could just be part of the -vanilla tasks. The depends_on: should still work, if they're generated as part of one matrix:

scripts/bsd/openbsd-prep-postgres.sh
```diff
@@ -0,0 +1,34 @@
...
+libxslt \
+lz4 \
+openpam \
+python%3.8 \
+readline \
+tcl%8.6 \
+\
+login_krb5 \
+openldap-client--gssapi \
+openldap-server--gssapi || true
```
anarazel, line 30: What's the deal with the `|| true` here? Won't that end up with the image build succeeding, even though there were package installation failures?

scripts/bsd/netbsd-prep-postgres.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+# Install required packages for running tests on netBSD
+PKG_PATH="http://cdn.NetBSD.org/pub/pkgsrc/packages/NetBSD/$(uname -p)/$(uname -r|cut -f '1 2' -d.)/All/" && \
+export PKG_PATH && \
+/usr/sbin/pkg_add pkgin && \
```
anarazel, line 6: Probably makes sense to install and configure pkgin in the base image.

scripts/bsd/netbsd-prep-postgres.sh
```diff
@@ -0,0 +1,30 @@
...
+/usr/pkg/bin/pkgin update && \
+mkdir /usr/pkg/pgsql && \
```
anarazel, line 8: Hm, what is /usr/pkg/pgsql about?

scripts/bsd/openbsd-prep-postgres.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+pkg_add -uvI && \
```
anarazel, line 3: This updates all the already installed packages, right? If so, that should probably be part of the base image script instead.

packer/bsd/openbsd_vanilla.pkr.hcl
```diff
@@ -0,0 +1,88 @@
...
+}
+
+source "qemu" "qemu-gce-builder" {
+
+  boot_command            = [
+    "S<enter><wait>",
+    "cat <<EOF >>install.conf<enter>",
+    "System hostname = openbsd70<enter>",
+    "Password for root = packer<enter>",
+    "Setup a user = packer<enter>",
```
anarazel, line 18: Is this user / password later deleted? We can't have that still be active in GCP...

packer/bsd/netbsd_vanilla.pkr.hcl
```diff
@@ -0,0 +1,113 @@
...
+  default = ""
+}
+
+source "qemu" "qemu-gce-builder" {
+
+  boot_command            = [
+      "<enter><wait6s>",
+      "<enter><wait6s>",
+      "<enter><wait6s>",
+      "<down><enter><wait6s>",
```
anarazel, line 17: Might be worth adding a few comments explaining what all these waits are about...

packer/bsd/netbsd_vanilla.pkr.hcl
```diff
@@ -0,0 +1,113 @@
...
+      "<enter><wait6s>",
+      "packer<enter><wait6s>",
+      "x<wait1s><enter><wait6s>",
+      "<enter><wait6s>",
+      "d<wait1s><enter><wait120s>",
+      "r<wait1s>o<wait1s>o<wait1s>t<wait1s>",
+      "<enter><wait6s>",
+      "p<wait1s>a<wait1s>c<wait1s>k<wait1s>e<wait1s>r<wait1s>",
+      "<enter><wait6s>",
+      "sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/g' /etc/ssh/sshd_config<enter><wait6s>",
```
anarazel, line 49: This should probably be reset at the end of the process. Or perhaps this could use ssh keys generated by packer?

packer/bsd/netbsd_vanilla.pkr.hcl
```diff
@@ -0,0 +1,113 @@
...
+      "echo '/sbin/dhcpcd -4' > /etc/rc.local<enter><wait1s>",
```
anarazel, line 50: This probably isn't the right method. Perhaps @macdice knows?

.cirrus.yml
nbyavuz: There are small differences like;

Existing packer: 
```    
packer build \
      -only "*.${CIRRUS_TASK_NAME}" \
      -timestamp-ui \
      -force \
      -var "image_date=$DATE" \
      -var gcp_project=$GCP_PROJECT \
      "$PACKERFILE"
```
OpenBSD / NetBSD:
```
    packer build \
      -timestamp-ui \
      -force \
      -var "image_date=$DATE" \
      -var "image_name=${CIRRUS_TASK_NAME}" \
      -var "bucket=$BUCKET" \
      -var gcp_project=$GCP_PROJECT \
      "$PACKERFILE"
```

Existing packer does not include
```
      -var "image_name=${CIRRUS_TASK_NAME}" \
      -var "bucket=$BUCKET" \
```
so, we need to add these variables to existing packer files, otherwise packer will complain about that. Also, their skip cases are different.

nbyavuz:
It seems like, I can not approve my PR(and I am not sure if my approve will trigger CI, since it says **owner** approval).

packer/bsd/netbsd_vanilla.pkr.hcl
nbyavuz: That is especially for forcing ipv4, otherwise NetBSD's ipv6 is not working correctly with QEMU. Installation of one packages takes around 30 minutes. Related issue: https://gitlab.com/qemu-project/qemu/-/issues/985

anarazel (commented):
Gah, forgot to submit those comments quite a while ago :(

gcp_project_setup.txt
```diff
@@ -26,6 +26,8 @@ gcloud iam service-accounts keys create \
 # FIXME: It'd be better to create a more restrictive role with just the
 # permissions we need, intead of using the pre-defined roles.
 
+# Create pg-ci-bucket for uploading images of openBSD and netBSD
+gsutil mb -p $GCP_PROJECT -b on gs://pg-ci-bucket
```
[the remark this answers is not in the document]
nbyavuz: Could you retry this? I was using `pg-ci-bucket` for dev project, renamed it as a `pg-ci-bucket-dev` now.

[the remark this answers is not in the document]
anarazel: Worked. Would still be good to add a note in the doc indicating that it needs to be unique. Perhaps we could just incorporate $GCP_PROJECT in the bucket name? Perhaps we name it $GCP_PROJECT-image-upload or such?

scripts/bsd/openbsd-prep-postgres.sh
```diff
@@ -0,0 +1,36 @@
...
+    login_krb5 \
+    openldap-client--gssapi \
+    openldap-server--gssapi
+
+# Set kernel parameters for running postgres tests
+echo "/sbin/sysctl kern.seminfo.semmni=2048" >> /etc/rc.local
+echo "/sbin/sysctl kern.seminfo.semmns=32768" >> /etc/rc.local
+
+# Forbid root username/password login
+sed -i 's/PermitRootLogin yes/PermitRootLogin prohibit-password/g' /etc/ssh/sshd_config
```
anarazel, line 36: Doesn't this mean the password is left in the 'vanilla' images? We shouldn't need password auth once the image is uploaded to GCP, since the accounts / ssh keys should be managed by the GCP scripts, no?

scripts/bsd/netbsd-prep-postgres.sh
```diff
@@ -0,0 +1,28 @@
...
+    ccache \
+    p5-IPC-Run \
+    flex \
+    pkgconf \
+    python38 \
+    icu \
+    lz4 \
+    libxslt \
+    tcl && \
+/usr/pkg/sbin/mozilla-rootcerts install
```
anarazel, line 21: The rootcerts should probably be installed as part of the vanilla image.

scripts/bsd/openbsd-prep-postgres.sh
nbyavuz: If 'we' want to connect the 'vanilla' images it works without passwords but packer needs to connect with username / password for creating 'postgres' images. We can add another step which is for creating second vanilla images from first vanilla images. We can remove password in these step, then we can remove first vanilla image but this is a bit costly.

anarazel: > If 'we' want to connect the 'vanilla' images it works without passwords but packer needs to connect with username / password for creating 'postgres' images

Why / When does packer need that? That step runs in gcp, and packer can create temporary ssh keys for use with gcp. That's how it works for the debian / freebsd images currently...

nbyavuz: These are the steps:

1. Create 'vanilla' VM from ISO file
2. Export that VM as a image file to the Google Images
    2.1. We have 'vanilla' image in Google Images now
3. Create and export 'postgres' image using 'vanilla' image

I think since we use vanilla VM's image(not raw image like in the GCP), packer can't connect the image without username and passwords. Packer can create temporary ssh keys but 'vanilla' image doesn't have public key of temporary ssh key so it can't connect with using ssh key pair.

Alternative idea: We can create 'postgres' image like we create 'vanilla' image. We can create it from ISO file, then no need to have a username / password.

anarazel: packer's ssh keys are supposed to be provided via the gcp infrastructure / scripts. If that doesn't work, something else is wrong. How exactly does it fail?

anarazel: Ping?

nbyavuz: This is what happens if I remove ssh's pasword 'postgres' image: https://cirrus-ci.com/task/4592026892107776

Startup script in these images work like that: search & download users' public keys to the user' folders. Then user need to pass private key to the ssh command. Since packer is not a user in our GCP, startup scripts won't download its ssh keys. I am not sure how packer provide its keys via the gcp infrastructure / scripts but if its something similar to metadata, then we can add that to startup scripts.

anarazel: I think it creates an ephemeral user. It might just be that it relies on `ssh_username            = "packer"` that the others use, i.e. that root doesn't work.

nbyavuz: Same error: https://cirrus-ci.com/task/4771415630020608

I can search details of how packer transfer its keys.

gcp_project_setup.txt
```diff
@@ -26,6 +26,9 @@ gcloud iam service-accounts keys create \
 # FIXME: It'd be better to create a more restrictive role with just the
 # permissions we need, intead of using the pre-defined roles.
 
+# Create bucket for uploading images of openBSD and netBSD, bucket name
+# should be unique
+gsutil mb -p $GCP_PROJECT-image-upload -b on gs://pg-ci-bucket
```
anarazel, line 31: How about including $GCP_PROJECT in the bucket name?

scripts/bsd/openbsd-prep-postgres.sh
anarazel: We create a user like that, that's why it works for linux / freebsd. I suspect that there's some problem with the script handling the key installation (likely not handling the unix username differing from the GCP username), preventing this from working.  Did you try doing this interactively with packer build -debug, and logging into the *bsd machine to see what happens on the machine?

nbyavuz: Done.

anarazel: What did you end up having to change?

gcp_project_setup.txt
```diff
@@ -26,6 +26,9 @@ gcloud iam service-accounts keys create \
 # FIXME: It'd be better to create a more restrictive role with just the
 # permissions we need, intead of using the pre-defined roles.
 
+# Create bucket for uploading images of openBSD and netBSD, bucket name
+# should be unique
+gsutil mb -p $GCP_PROJECT-bucket -b on gs://pg-ci-bucket
```
anarazel, line 31: This doesn't look right? Think you meant to change the end of the line, not the project name?

anarazel:
Could you rebase this? The sid* failures should be fixed there...

gcp_project_setup.txt
nbyavuz: My bad, I will update it.

files/bsd/rc.local.sh
```diff
@@ -0,0 +1,143 @@
...
+        echo "Hostname is correct"
+    fi
+else
+    echo "Could not discover hostname"
+fi
+
+### set up user's keys ###
+
+instance_keys=$(curl -s -H "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/project/attributes/ssh-keys)
+packer_temp_root_key=$(curl -s -H "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/instance/attributes/ssh-keys)
```
anarazel, line 64: Hm - it doesn't seem quite right to reference packer here, it's not packer specific. And the instance_keys name, in the previous line, seems a bit misleading too, given that it's fetch project level keys? Perhaps worth a a comment or two about fetching project and instance level keys?

anarazel (approved):
Yay!

PR #11, opened 2022-06-16
Clear users and ssh keys before publishing image

nbyavuz:
I realized that since we are creating postgres images from vanilla images, vanilla images' rc.local script will create users and these users will exist in postgres images. We need to remove these users. This PR aims to fix that problem.

### DIFF
--- a/files/bsd/rc.local.sh
+++ b/files/bsd/rc.local.sh
@@ -11,7 +11,7 @@ then
     {
         /etc/rc.d/syslogd restart > /dev/null
     }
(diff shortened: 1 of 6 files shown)


anarazel:
Shouldn't this be done for all images, rather than just the BSD ones?

nbyavuz:
> Shouldn't this be done for all images, rather than just the BSD ones?

Yes, creating issue and assigning to myself.

anarazel (approved):

PR #18, opened 2022-08-01
Ci fixes

### LINKED ISSUE #13: concurrent image builds fail due to instance name conflicts
If two images are built at the same time, they currently will often fail due to the CI image names conflicting. That's because I (andres) made the image names predictable, to avoid a lot of instances building up in case of failure. But it's likely more confusing than it's worth.  Perhaps we can give instances a shorter max lifetime? And then name them with a recognizable prefix, but a random postfix?

### LINKED ISSUE #14: reduce redundant references to freebsd version number
As seen in 3c43073ff4a48790aba59039f3c4a9fbb969bc63 there's multiple references to freebsd-13-$n in the packer file. It'd be good to reduce that to exactly one.

nbyavuz:
Fixes #13, #14.

### DIFF
--- a/packer/freebsd.pkr.hcl
+++ b/packer/freebsd.pkr.hcl
@@ -1,5 +1,6 @@
 variable "image_date" { type = string }
 variable "gcp_project" { type = string }
+variable "task_name" { type = string }
 
 variable "prefix" {
   type = string
@@ -11,7 +12,7 @@ locals {
 
   freebsd_gcp_images = [
     {
-      name = "freebsd-13"
+      name = "${var.task_name}"
       zone = "us-west1-a"
       machine = "e2-highcpu-4"
     },
--- a/packer/freebsd.pkr.hcl
+++ b/packer/freebsd.pkr.hcl
@@ -49,7 +49,7 @@ build {
 
       zone = tag.value.zone
       machine_type = tag.value.machine
-      instance_name = "build-${local.name}-${tag.value.name}"
+      instance_name = "build-${tag.value.name}-${var.image_date}"
     }
   }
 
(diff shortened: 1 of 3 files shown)


packer/freebsd.pkr.hcl
```diff
@@ -48,7 +49,7 @@ build {
 
       zone = tag.value.zone
       machine_type = tag.value.machine
-      instance_name = "build-${local.name}-${tag.value.name}"
+      instance_name = "build-${local.name}-${tag.value.name}-${var.image_date}"
```
anarazel, line 52: What format does image_date have? It seems this either could lead to duplicates if it really is just the date, or to too long instance names (IIRC they're limited to 63 chars or such)

anarazel: > It is the same DATE variable which is used in image names and it is like 2022-08-01t14-20

Ok. I guess it'd be better to add seconds for this purpose?

> For now, the longest name of the instance is build-pg-ci-sid-newkernel-uring-2022-08-01t14-20 which is 48 chars.

Perhaps we could remove "pg-ci-" from the image name?

nbyavuz: > > It is the same DATE variable which is used in image names and it is like 2022-08-01t14-20
> 
> Ok. I guess it'd be better to add seconds for this purpose?

Yes, that could work.

> 
> > For now, the longest name of the instance is build-pg-ci-sid-newkernel-uring-2022-08-01t14-20 which is 48 chars.
> 
> Perhaps we could remove "pg-ci-" from the image name?

Are we removing "pg-ci" from image names or instance names?

anarazel: > Are we removing "pg-ci" from image names or instance names?

err, instance, not image names.

nbyavuz: Done.

anarazel (approved):

PR #20, opened 2022-08-01
Refactoring net/openbsd

nbyavuz:
Refactoring net/openbsd due to recent changes.

### DIFF
--- a/packer/bsd/netbsd_postgres.pkr.hcl
+++ b/packer/bsd/netbsd_postgres.pkr.hcl
@@ -10,15 +10,21 @@ variable "prefix" {
   default = ""
 }
 
+locals {
+  name = "${var.prefix}pg-ci"
+  version = "9-2"
+}
+
 source "googlecompute" "netbsd-vanilla" {
   disk_size               = "25"
   disk_type               = "pd-ssd"
   preemptible             = "true"
   project_id              = var.gcp_project
-  source_image_family     = "pg-ci-netbsd-9-2-vanilla"
+  source_image_family     = "${local.name}-netbsd-${local.version}-vanilla"
   source_image_project_id = ["${var.gcp_project}"]
-  image_family            = "pg-ci-${var.image_name}"
-  image_name              = "${var.prefix}pg-ci-${var.image_name}-${var.image_date}"
+  image_family            = "${local.name}-${var.image_name}"
+  image_name              = "${local.name}-${var.image_name}-${var.image_date}"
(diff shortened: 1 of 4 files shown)


packer/bsd/openbsd_vanilla.pkr.hcl
```diff
@@ -42,6 +46,9 @@ source "qemu" "qemu-gce-builder" {
   format                  = "raw"
   vm_name                 = "disk.raw"
   output_directory        = "output"
+  qemuargs                = [
+    [ "-serial", "stdio" ]
```
anarazel, line 50: Looks like that didn't work, neither net/openbsd shows more details. I suspect the reason we don't see more is because you didn't export PACKER_LOG:
https://github.com/anarazel/pg-vm-images/blob/main/.cirrus.yml#L168
that way it won't be visible to packer afaics.

nbyavuz: Yes, my bad. It works now.

packer/bsd/netbsd_postgres.pkr.hcl
```diff
@@ -10,23 +10,28 @@ variable "prefix" {
   default = ""
 }
 
+locals {
+  name = "${var.prefix}pg-ci"
+}
```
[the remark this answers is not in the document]
nbyavuz: Done.

packer/bsd/openbsd_vanilla.pkr.hcl
anarazel: I think we should split the serial console stuff into its own PR. The rest here we can apply now, but since the serial stuff isn't yet working properly...

packer/bsd/netbsd_postgres.pkr.hcl
```diff
@@ -10,23 +10,29 @@ variable "prefix" {
   default = ""
 }
 
+locals {
+  name = "${var.prefix}pg-ci"
+  version = "9-2"
+}
+
```
anarazel, line 17: Wonder if, as part of a separate change, we should add a common variable file for -vanilla and -postgres, which then could define the versions (including image URL & hash) in one place .

anarazel (approved):

packer/bsd/netbsd_postgres.pkr.hcl
nbyavuz: That makes sense, creating an issue.

PR #27, opened 2022-08-22
net/openbsd: merge packer files

### LINKED ISSUE #22: Create common variable file for net/openbsd
We can create a common variable file for defining versions, image URLS, hashes etc. in one place.

nbyavuz:
Fixes #22.

This PR merges net/openbsd's packer files and creates one common variable for each net/openbsd.

Reasoning is:
- When there are common variable files for net/openbsd, packer complains about undefined variables in the .hcl files. Same thing happens when there are separate variable files for net/openbsd and separate vanilla/postgres.hcl files.

- Also, both bsd's .hcl files are very similar(both vanilla and postgres files). When something is to be added, almost all time same feature is added the other one. Combining these files will increase maintainability in the future.


nbyavuz:
> Oh, I wasn't actually imagining to merge the whole files. Just that we'd use a common file for defining variables, using packer build -var-file=... - what made you decide to combine the files as a whole?

1 - When we have common variable file for net/openbsd, packer complains about not defined variables in the .hcl files. I didn't want to define each other's variables in the other's .hcl file to bypass that error. The same goes for having separate variable files for net/openbsd and separate vanilla/postgres.hcl files.
2 - Both bsd's .hcl files are very similar(both vanilla and postgres files). When we were adding something, almost all time we are adding same feature the other one. I thought combining them will increase maintainability in the future.

anarazel:
Such reasoning should be in the commit message :)

nbyavuz:
> Such reasoning should be in the commit message :)

Yes, you are right. I added these descriptions to the commit message.

.cirrus.yml
```diff
@@ -127,29 +127,24 @@ task:
...
+            - netbsd-vanilla
 
-    # see openbsd case
-    - name: netbsd-9-postgres
-      depends_on:
-        - netbsd-9-2-vanilla
-      env:
-        PACKERFILE: packer/bsd/netbsd_postgres.pkr.hcl
+  env:
+    PACKERFILE: packer/bsd/bsd.pkr.hcl
```
anarazel, line 148: The only difference to the other packer invocation now is that -var bucket  is defined for net/openbsd, but not for the others. Could we, possibly in a separate PR, combine the packer invocations?

packer/bsd/bsd.pkr.hcl
```diff
@@ -97,11 +49,11 @@ source "qemu" "qemu-gce-builder" {
 }
 
 build {
-  name="netbsd-vanilla"
+  name = "vanilla"
   sources = ["source.qemu.qemu-gce-builder"]
 
   provisioner "shell" {
-    script = "scripts/bsd/netbsd-prep-gce.sh"
+    script = "${var.gce_script}"
```
anarazel, line 56: I think it might look nicer to address these with `only = ["whatever-netbsd-task's-name-is"]`?

packer/bsd/bsd.pkr.hcl
```diff
@@ -2,6 +2,15 @@ variable "bucket" { type = string }
 variable "gcp_project" { type = string }
```
anarazel, line 2: I think there's not really a need for the packer/bsd/ subdirectory anymore?

.cirrus.yml
```diff
@@ -166,11 +161,14 @@ task:
...
   build_image_script: |
     DATE=$(date --utc +'%Y-%m-%dt%H-%M-%S')
+    PKRVARFILE=packer/bsd/${CIRRUS_TASK_NAME%-*}.pkrvars.hcl
     export PACKER_LOG=1
 
     packer build \
+      -only "${BUILD_TYPE}.*" \
       -timestamp-ui \
       -force \
+      -var-file="${PKRVARFILE}" \
```
anarazel, line 171: Could you adjust the top-level Makefile as well? I use that to locally test that packer file is at least basically sound, with `make pre-commit`.

packer/bsd/bsd.pkr.hcl
nbyavuz: Yes, moved them.

packer/bsd/bsd.pkr.hcl
nbyavuz: Isn't setting name for build and calling that build with -only enough?

anarazel: What I mean is that instead of using a variable for the script, you can do something like

```hcl
  provisioner "shell" {
    script = "scripts/bsd/netbsd-prep-gce.sh"
    only = ['netbsd'] #actually a different name here
  }
  provisioner "shell" {
    script = "scripts/bsd/openbsd-prep-gce.sh"
    only = ['openbsd'] #actually a different name here
  }
```
This isn't quite right, but should get the idea across.

nbyavuz: Done.

packer/netbsd_openbsd.pkr.hcl
```diff
@@ -137,7 +103,7 @@ build {
   # lock root because we don't want that root is accessible by using
   # username-password, root is still reachable from ssh key
   provisioner "shell" {
-    inline = ["/usr/sbin/usermod -C yes root"]
+    inline = ["${var.lock_root_command}"]
   }
```
anarazel, line 107: I'd do the same as the only stuff above here.

packer/netbsd_openbsd.pkr.hcl
```diff
@@ -97,11 +49,25 @@ source "qemu" "qemu-gce-builder" {
...
+
+  dynamic "source" {
+    for_each = var.vanilla_name
+    labels = ["source.qemu.qemu-gce-builder"]
+    iterator = tag
+
+    content {
+      name = tag.value.name
+    }
+  }
```
anarazel, line 62: What is this for?

nbyavuz: This is mainly for using `-only "*.${CIRRUS_TASK_NAME}"` while building net/openbsd images, so we can easily combine the packer invocations later. Like you mentioned in .cirrus.yml, we can't use `var or local for` build.name so I implemented it in this way.

anarazel (approved):

PR #28, opened 2022-09-21
windows: remove files we don't need

### LINKED ISSUE #1: windows: reduce image size by removing unnecessary files from debugger installation
The windows instances start slow because the containers need to be downloaded. There's plenty in those containers that we don't actually need. There's e.g. several hundred MB of .pdb and arm specific files. While not normally a good idea, for a dedicated container like this we should just remove them.

CC: @justinpryzby

nbyavuz:
Improves #1

### DIFF
--- a/scripts/windows_install_vs_2019.ps1
+++ b/scripts/windows_install_vs_2019.ps1
@@ -17,6 +17,19 @@ Start-Process -Wait `
       '--add', 'Microsoft.VisualStudio.Component.Windows10SDK.20348'
 if (!$?) { throw 'cmdfail' }
 
+# clear pdb files to download them later
+du -shc "C:\BuildTools\VC\Tools\MSVC"
+rm "C:\BuildTools\VC\Tools\MSVC\*\lib\**\**\*pdb"
+du -shc "C:\BuildTools\VC\Tools\MSVC"
+
+# clear not required arm files
+du -shc "" "C:\Program Files (x86)\Windows Kits"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\ucrt\arm"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\ucrt\arm64"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\um\arm"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\um\arm64"
+du -shc "" "C:\Program Files (x86)\Windows Kits"
+
 cd c:\
 Remove-Item C:\t -Force -Recurse
 Remove-Item -Force -Recurse ${Env:TEMP}\*


scripts/windows_install_vs_2019.ps1
```diff
@@ -17,6 +17,19 @@ Start-Process -Wait `
...
+rm "C:\BuildTools\VC\Tools\MSVC\*\lib\**\**\*pdb"
+du -shc "C:\BuildTools\VC\Tools\MSVC"
+
+# clear not required arm files
+du -shc "" "C:\Program Files (x86)\Windows Kits"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\ucrt\arm"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\ucrt\arm64"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\um\arm"
+rm -r "C:\Program Files (x86)\Windows Kits\10\Lib\*\um\arm64"
+du -shc "" "C:\Program Files (x86)\Windows Kits"
```
anarazel, line 31: Hm, do these files actually get added in this layer?

anarazel: Ping?

nbyavuz: Yes, double checked. Inside of windows_ci_vs_base:

```
PS C:\> ls


    Directory: C:\


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----        9/21/2022   1:24 PM                icu
d-----        9/21/2022   1:24 PM                lz4
d-----        9/20/2022  12:15 PM                openssl
d-r---        6/23/2022   9:53 PM                Program Files
d-----         6/9/2022  10:40 PM                Program Files (x86)
d-----        9/20/2022  12:15 PM                python
d-----        9/20/2022  12:15 PM                strawberry
d-r---         6/9/2022  10:45 PM                Users
d-----        9/20/2022  12:14 PM                Windows
d-----        9/20/2022  12:14 PM                Windows Kits
d-----        9/21/2022   1:24 PM                winflexbison
d-----        9/21/2022   1:24 PM                zlib
d-----        9/21/2022   1:24 PM                zstd
-a----        4/22/2022   3:57 AM           5510 License.txt

PS C:\>
```

```
PS C:\Program Files (x86)> ls


    Directory: C:\Program Files (x86)


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----        9/15/2018  10:21 AM                common files
d-----        9/15/2018  10:12 AM                internet explorer
d-----        9/15/2018  10:12 AM                Microsoft.NET
d-----        9/15/2018  10:12 AM                WindowsPowerShell


PS C:\Program Files (x86)>
```

anarazel: The reason I was asking is that the debugger is installed in "Windows Kits" as well - but in "C:/Windows Kits"

anarazel: Would be good to check if there's any duplication between the two directories...

PR #32, opened 2022-11-09
Use windows vms instead of containers

### LINKED ISSUE #31: windows: Try to move to windows VMs instead of containers
Right now we use containers to run tests on windows. But they're slow to start (because the container image gets pulled every time) and slow to run (I suspect due to containerization and potentially anti-virus overhead).

I'd like to convert our windows dockerfiles to packer. First using packer to continue generating windows containers. But then subsequently we can generate windows VM images instead. cirrus-ci can use windows VMs as well. Then we can compare performance between windows VMs and windows containers.

nbyavuz:
Fixes #31.


scripts/windows_install_openssl.ps1
```diff
@@ -13,7 +13,7 @@ $ErrorActionPreference = "Stop"
 $filepath = "$Env:TEMP/openssl-setup.exe"
 
 echo "downloading openssl"
-curl.exe -o "$filepath" -sSL https://slproweb.com/download/Win64OpenSSL-1_1_1q.exe
+curl.exe -o "$filepath" -sSL https://slproweb.com/download/Win64OpenSSL-1_1_1s.exe
```
anarazel, line 16: Can you open a separate PR for this? We need to merge it soon.

packer/windows.pkr.hcl
```diff
@@ -0,0 +1,154 @@
...
+  name = "${var.prefix}pg-ci"
+
+  windows_gcp_images = [
+    {
+      name = "${var.task_name}"
+    },
+  ]
+}
+
+source "googlecompute" "windows-vm" {
```
anarazel, line 20: Have you tried to generate docker containers using packer as well? I'd like to have a somewhat orderly transition for CI, and that'll be harder if we just stop producing the containers on a certain date.

.cirrus.yml
```diff
@@ -347,12 +309,16 @@ task:
     - bison --version
     - flex --version
 
+
 task:
   name: test-windows-container-mingw64-amd64
```
anarazel, line 314: The name is wrong now, not using a container anymore :)

packer/windows.pkr.hcl
```diff
@@ -0,0 +1,154 @@
...
+    labels = ["source.googlecompute.windows-vm"]
+    iterator = tag
+
+    content {
+      name = tag.value.name
+    }
+  }
+
+
+  # FIXME: is there a way to copy only windows* scripts?
```
anarazel, line 56: Why do we need to copy them at all? Can't we just run them with a script provisioner?

packer/windows.pkr.hcl
```diff
@@ -0,0 +1,154 @@
...
+  zone                    = "us-west1-a"
+  machine_type            = "n2-standard-4"
+  instance_name           = "build-${var.task_name}-${var.image_date}"
+  communicator            = "winrm"
+  winrm_username          = "packer_user"
+  winrm_insecure          = true
+  winrm_use_ssl           = true
+  winrm_timeout           = "10m"
+  metadata = {
+    windows-startup-script-cmd = "winrm quickconfig -quiet & net user /add packer_user & net localgroup administrators packer_user /add & winrm set winrm/config/service/auth @{Basic=\"true\"}"
```
anarazel, line 36: Do we need to remove that user?

scripts/windows_install_openssl.ps1
nbyavuz: PR Link: https://github.com/anarazel/pg-vm-images/pull/33

packer/windows.pkr.hcl
nbyavuz: I am not sure if that is possible, because we are connecting to the image as this user. Also, [Cirrus images have this user](https://github.com/cirruslabs/vm-images/blob/master/googlecompute/windows_images.json#L21), too.

packer/windows.pkr.hcl
nbyavuz: I didn't try it yet, will work on that.

nbyavuz:
> I think this can be closed now?

Yes, closed.

PR #38, opened 2022-11-17
Trigger CI if pkrvars files are changed

nbyavuz:


anarazel:
We can improve this if we end up with more variable files, but for now this seems good enough. Thanks!

PR #39, opened 2022-11-20
Add doc on how to use BSD images.

xiaq:


anarazel (approved):
@nbyavuz Perhaps worth adding the other images as well?

anarazel:
Thanks!

PR #46, opened 2023-01-05
Add IMAGE_IDENTITY environment variable

### LINKED ISSUE #45: Add image version / name as an environment variable to generated contains / VMs
It'd be good if all images had a IMAGE_IDENTITY (or such) environment variable containing the name + date of an image. That way we could output that for CI runs, and make it easier to pinpoint when an issue was introduced in a specific image version. 

It'd be good if IMAGE_IDENTITY would allow to specify the specific image (e.g. for VMs it should be the image name, so it could be used by replacing image_family: $basename, with image: $IMAGE_IDENTITY from a previous run).

nbyavuz:
Fixes #45.

### DIFF
--- a/files/bsd/rc.local.sh
+++ b/files/bsd/rc.local.sh
@@ -2,6 +2,10 @@
 
 # Save to /etc/rc.local, to be ran on startup
 
+# Set IMAGE_IDENTITY variable
+. /etc/environment.d/image_identity.conf
+export IMAGE_IDENTITY
+
 if [ "$(uname)" = "NetBSD" ]
 then
     export PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/pkg/bin:/usr/pkg/sbin
--- a/packer/freebsd.pkr.hcl
+++ b/packer/freebsd.pkr.hcl
@@ -9,6 +9,7 @@ variable "prefix" {
 
 locals {
   name = "${var.prefix}pg-ci"
+  image_identity = "${local.name}-${var.task_name}-${var.image_date}"
 
   freebsd_gcp_images = [
     {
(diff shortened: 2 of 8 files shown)


anarazel:
For the linux VMs you can create a file in /etc/environment.d/.

files/bsd/rc.local.sh
```diff
@@ -2,6 +2,10 @@
 
 # Save to /etc/rc.local, to be ran on startup
 
+# Set IMAGE_IDENTITY variable
+. /etc/environment.d/image_identity.conf
+export IMAGE_IDENTITY=${IMAGE_IDENTITY}
```
anarazel, line 7: I think "export IMAGE_IDENTITY" should suffice.

packer/freebsd.pkr.hcl
```diff
@@ -9,6 +9,7 @@ variable "prefix" {
 
 locals {
   name = "${var.prefix}pg-ci"
+  image_name = "${local.name}-${var.task_name}-${var.image_date}"
```
anarazel, line 12: In other places you treat image_name as the portion without the date. Perhaps it'd be better to name the variable image_identity?

packer/netbsd_openbsd.pkr.hcl
```diff
@@ -141,7 +150,7 @@ source "googlecompute" "postgres" {
   source_image_family     = "${local.name}-${var.name}-vanilla-${var.version}"
   source_image_project_id = ["${var.gcp_project}"]
   image_family            = "${local.name}-${var.image_name}-${var.version}"
-  image_name              = "${local.name}-${var.image_name}-${var.image_date}"
+  image_name              = local.image_identity
```
anarazel, line 153: Huh, why do we use the version in the image family, but not the name? That seems wrong.

anarazel (commented):
Does make pre-commit need any updates?

Wondering about that made realize that windows packer isn't tested there.

packer/netbsd_openbsd.pkr.hcl
nbyavuz: I added version to image_name. I think we should kept version in the family but maybe only major version? 

Also, I moved `gcloud compute images add-iam-policy-binding` part from `.cirrus.yml` to `netbsd_openbsd.pkr.hcl` because version is got from pkrvars files, not from task_name and it worked. Should we apply same for other tasks?

anarazel: > I added version to image_name. I think we should kept version in the family but maybe only major version?

Yes, that'd make sense. But let's do that in a separate PR.

> Also, I moved gcloud compute images add-iam-policy-binding part from .cirrus.yml to netbsd_openbsd.pkr.hcl because version is got from pkrvars files, not from task_name and it worked. Should we apply same for other tasks?

Hm. It somehow seems wrong to do the iam-policy stuff inside packer, but I can't really explain why. However, you're right, the names don't match up right now. Either way, could we do that as part of a separate PR?

anarazel (approved):
Looks good. Feel free to merge - I thought you might want to merge the IPC::Run one first, so I didn't merge it just now.

PR #47, opened 2023-01-23
ci: Fix packer does not stop on powershell errors

nbyavuz:
Adding `$ErrorActionPreference = 'Stop'` is solving VM issue but when it is added to container, it doesn't run scripts. So, I used `Invoke-Expression` for containers. I tried using wrong commands, exit with other than 0, throwing errors and all of them stopped the execution.

For the containers:

We need to use `$ErrorActionPreference = 'Stop'` in inline scripts too because packer is putting these inline commands to the *.ps1; then it executes these files.

But here could be some problems(?):

* Only `Invoke-Expression` worked correctly but there are some issues about that: 

> [Take reasonable precautions when using the Invoke-Expression cmdlet in scripts. When using Invoke-Expression to run a command that the user enters, verify that the command is safe to run before running it. In general, it is best to design your script with predefined input options, rather than allowing freeform input.](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-expression?view=powershell-7.3)

* When passing `Invoke-Expression -ErrorAction Stop -Command '. {{.Vars}};. {{.Path}}; exit $LastExitCode'` from .cirrus.yml file, Cirrus always calculates `$LastExitCode` as an environment variable and calls script with `Invoke-Expression -ErrorAction Stop -Command '. {{.Vars}};. {{.Path}}; exit '`. I couldn't escape `$LastExitCode`, so I put this command to var-file instead of .cirrus.yml.

### DIFF
--- a/packer/windows.pkr.hcl
+++ b/packer/windows.pkr.hcl
@@ -10,6 +10,15 @@ variable "docker_repo" {
   default = ""
 }
 
+# Execute command for VM instances, containers will overwrite this because
+# docker builder uses docker exec to run commands and that is hardcoded.
+# It doesn't capture errors correctly when VM's execute command is used.
+# So, containers have another execute_command.
+variable "execute_command" {
+  type = string
+  default = "powershell -executionpolicy bypass \"& { if (Test-Path variable:global:ProgressPreference){$ProgressPreference='SilentlyContinue'}; $ErrorActionPreference = 'Stop' ;. {{.Vars}}; &'{{.Path}}'; exit $LastExitCode }\""
+}
+
 variable "docker_server" {
   type = string
   default = ""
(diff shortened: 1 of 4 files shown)


packer/windows.pkr.hcl
```diff
@@ -10,6 +10,12 @@ variable "docker_repo" {
   default = ""
 }
 
+variable "execute_command" {
+  type = string
+  # execute command for vm instances, containers will overwrite this
+  default = "powershell -executionpolicy bypass \"& { if (Test-Path variable:global:ProgressPreference){$ProgressPreference='SilentlyContinue'}; $ErrorActionPreference = 'Stop' ;. {{.Vars}}; &'{{.Path}}'; exit $LastExitCode }\""
```
anarazel, line 16: Explain why we are doing this, and particularly why we are overwriting it for docker containers.

anarazel (approved):
Except for that one comment I just made about commenting, I think it looks. You need to remove the "draft" status though ...

PR #48, opened 2023-02-03
Fix IPC:Run installation on MinGW

nbyavuz:
MinGW was installing both `make-4.4-1` and `mingw-w64-ucrt-x86_64-make-4.4-2` packages before, but it started not to install `mingw-w64-ucrt-x86_64-make-4.4-2` (I am not sure why). I added that package. 

Also, Windows VM installation's `skip: !changesInclude` part didn't have `scripts/windows*` but when this is added; linux images' installation will be triggered too. I refactored that part.

CI Run after fix applied: https://cirrus-ci.com/task/6566742164504576

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -87,6 +87,7 @@ task:
 
     - env:
         PACKERFILE: packer/linux_debian.pkr.hcl
+        SCRIPTS: scripts/linux_debian_*
 
       matrix:
         - name: bullseye
@@ -96,6 +97,7 @@ task:
 
     - env:
         PACKERFILE: packer/windows.pkr.hcl
+        SCRIPTS: scripts/windows*
 
       matrix:
         - name: windows-ci-vs-2019
@@ -106,7 +108,7 @@ task:
     cpu: 0.5
     memory: 256Mi
 
-  skip: $CIRRUS_LAST_GREEN_CHANGE != '' && $CIRRUS_CRON != 'regular-rebuild' && !changesInclude('.cirrus.yml', 'docker/linux_debian_packer', 'scripts/linux_debian_*', $PACKERFILE)
+  skip: $CIRRUS_LAST_GREEN_CHANGE != '' && $CIRRUS_CRON != 'regular-rebuild' && !changesInclude('.cirrus.yml', 'docker/linux_debian_packer', $SCRIPTS, $PACKERFILE)
   auto_cancellation: false
 
   <<: *gcp_auth_unix
--- a/scripts/windows_install_mingw64.ps1
+++ b/scripts/windows_install_mingw64.ps1
@@ -19,9 +19,12 @@ msys 'pacman --noconfirm -Scc' ;
 
 echo 'installing packages' ;
 msys 'pacman -S --needed --noconfirm git bison flex make diffutils \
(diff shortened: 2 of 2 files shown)


anarazel:
Why did the lack of make lead to missing IPC::Run, without failing during installation?

anarazel:
> Also, Windows VM installation's skip: !changesInclude part didn't have scripts/windows* but when this is added; linux images' installation will be triggered too. I refactored that part.

You might not be surprised by this: It's imo much better to do stuff like this separately.

.cirrus.yml
```diff
@@ -81,32 +81,38 @@ gcp_docker_auth_win: &gcp_docker_auth_win
...
-        - name: sid-newkernel
-        - name: sid-newkernel-uring
-
-    - env:
-        PACKERFILE: packer/windows.pkr.hcl
-
+        - name: freebsd-13
+          env:
+            PACKERFILE: packer/freebsd.pkr.hcl
+            SCRIPTS: scripts/linux_debian_*
```
anarazel, line 89: Freebsd uses linux_debian_*?

And I don't think you actually use SCRIPTS anywhere?

.cirrus.yml
```diff
@@ -81,32 +81,38 @@ gcp_docker_auth_win: &gcp_docker_auth_win
 
 task:
   matrix:
-    - name: freebsd-13
-      env:
-        PACKERFILE: packer/freebsd.pkr.hcl
```
anarazel, line 86: I don't really get why you moved freebsd into a different level of the matrix?

nbyavuz:
> Why did the lack of make lead to a missing IPC::Run, without failing during installation?

I think it is about `msys '(echo; echo o conf recommends_policy 0; echo notest install IPC::Run) | cpan'` line, make is needed for installing IPC::Run but this pipeline didn't fail although it logs:
```
Failed during this command:
JDB/Win32-Process-0.17.tar.gz                : make NO
HAARG/Win32-ShellQuote-0.003001.tar.gz       : make NO
TODDR/IPC-Run-20220807.0.tar.gz              : make NO

cpan[3]> Warning: Configuration not saved.
Lockfile removed.
```
Instead of this method, I installed [cpanm](https://www.cpan.org/modules/INSTALL.html) and installed IPC:Run with using cpanm. I tested that it failed when make is missing.

scripts/windows_install_mingw64.ps1
```diff
@@ -19,9 +19,17 @@ msys 'pacman --noconfirm -Scc' ;
...
 
-# Install perl modules to enable tap tests
+### Install perl modules to enable tap tests
 msys 'where perl'
-msys '(echo; echo o conf recommends_policy 0; echo notest install IPC::Run) | cpan'
+
+echo 'Install cpanm'
+# Autoconfigure first
+msys 'echo | cpan'
+msys 'cpan App::cpanminus'
```
anarazel, line 31: Well, now this can fail to install without erroring out :)

I'd probably just leave it as before, but just add a `perl -mIPC::Run -e 1` as a separate command - that'll fail if IPC::Run isn't installed.

anarazel (commented):
Unfortunately the image build is failing for an independent reason. Trying to fix that.

anarazel:
> Unfortunately the image build is failing for an independent reason. Trying to fix that.

I fixed the failure on bullseye. The various sid images should fix themselves in a few hours, we can ignore that here.

anarazel (approved):
Looks good. Feel free to merge.

Thanks for fixing this!

PR #49, opened 2023-02-07
linux: Don't fail if google-cloud-sdk isn't installed

anarazel:
For unknown reasons occasionally the source image doesn't contain google-cloud-sdk, causing apt-get purge to fail, which in turn causes packer to fail. Just accept an error while purging google-cloud-sdk.

### DIFF
--- a/packer/linux_debian.pkr.hcl
+++ b/packer/linux_debian.pkr.hcl
@@ -93,9 +93,11 @@ build {
 
         # Remove unnecessary packages, to reduce image size
         apt-get purge -y \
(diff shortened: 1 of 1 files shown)


anarazel:
The failure of sid to build is transient, caused by libldap-2.5-0:amd64 being newer than libldap-2.5-0:i386, because the new version of libldap-2.5-0:i386 hasn't yet been processed at https://incoming.debian.org/debian-buildd/dists/buildd-unstable/main/binary-i386/

PR #51, opened 2023-02-10
ci: Refactor task naming logic

nbyavuz:


.cirrus.yml
```diff
@@ -95,13 +95,17 @@ task:
         - name: sid-newkernel
         - name: sid-newkernel-uring
 
-    - env:
+    - name: windows-ci-vs-2019-VM
+      env:
         PACKERFILE: packer/windows.pkr.hcl
         SCRIPTS: scripts/windows*
+        TASK_NAME: windows-ci-vs-2019
```
anarazel, line 102: Maybe s/TASK/IMAGE/? Since that's really what we're configuring?

.cirrus.yml
```diff
@@ -119,19 +123,25 @@ task:
   build_image_script: |
     DATE=$(date --utc +'%Y-%m-%dt%H-%M-%S')
 
+    # if TASK_NAME is not set use CIRRUS_TASK_NAME instead
+    if [[ ! -v TASK_NAME ]]; then
+      TASK_NAME=${CIRRUS_TASK_NAME}
+    fi
+    echo "TASK_NAME = ${TASK_NAME}"
```
anarazel, line 130: Why don't we just migrate everything over? It's a small enough file that I think "backward compatibility" hacks shouldn't be needed.

.cirrus.yml
```diff
@@ -95,13 +95,17 @@ task:
...
+        TASK_NAME: windows-ci-vs-2019
 
-      matrix:
-        - name: windows-ci-vs-2019
-        - name: windows-ci-mingw64
+    - name: windows-ci-mingw64-VM
+      env:
+        PACKERFILE: packer/windows.pkr.hcl
+        SCRIPTS: scripts/windows*
+        TASK_NAME: windows-ci-mingw64
```
anarazel, line 108: I was imagining that we wouldn't set 'name' directly anymore, but instead have something like this:
```yaml
name: Building VM Image: $IMAGE_NAME
name: Building Docker Image: $IMAGE_NAME
```
in the respective tasks.

nbyavuz: Then depend_on should change. I created a draft PR to see changes: https://github.com/anarazel/pg-vm-images/pull/54

PR #53, opened 2023-02-13
ci: Deprecate images on failure

nbyavuz:


.cirrus.yml
```diff
@@ -79,6 +79,11 @@ gcp_docker_auth_win: &gcp_docker_auth_win
     - gcloud auth configure-docker --quiet %GCP_REGION%-docker.pkg.dev
 
 
+on_failure_googlecompute: &on_failure_googlecompute
```
anarazel, line 82: Maybe add a comment saying something like "Occasionally image creation succeeds despite signalling a failure. Without further action that'd result in accesses to the image failing due to lacking privileges. Deprecating the image avoids it being used by default, but leaves the image around for potential investigation."

packer/windows.pkr.hcl
```diff
@@ -76,6 +76,7 @@ source "googlecompute" "windows" {
   winrm_insecure          = true
   winrm_use_ssl           = true
   winrm_timeout           = "10m"
+  state_timeout           = "10m"
```
anarazel, line 79: Hm. Do you suspect this will actually help? 

Seems to be consistently taking ~3min in the ~5 cases runs I looked at.

I'd normally prefer to do something like this in a PR with a different title than "ci: Deprecate images on failure" :)

.cirrus.yml
```diff
@@ -153,6 +165,8 @@ task:
         - name: netbsd-postgres
           depends_on:
             - netbsd-vanilla
+          on_failure:
+            <<: *on_failure_googlecompute
```
anarazel, line 170: Why don't we just do this at the task: level, instead of repeating it in multiple matrices?

.cirrus.yml
```diff
@@ -167,8 +181,11 @@ task:
   info_script:
     - packer --version
 
-  build_image_script: |
+  export_date_script: |
     DATE=$(date --utc +'%Y-%m-%dt%H-%M-%S')
+    echo "DATE=${DATE}" | tee -a $CIRRUS_ENV
+
+  build_image_script: |
```
anarazel, line 188: I previously was thinking that it'd be nice to have the same timestamp for all images. But unfortunately it doesn't look like there's a proper CIRRUS_* variable for it. Perhaps we should suggest that as a feature?

nbyavuz: Yes, getting creation time of the task could be very helpful.

anarazel: Yes, please do.

anarazel (approved):

PR #54, opened 2023-02-15
ci: Refactor task naming logic - 2

nbyavuz:
This PR is created for discussing task_name changes.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,6 +5,9 @@ only_if: $CIRRUS_REPO_OWNER == 'anarazel' && $CIRRUS_REPO_NAME == 'pg-vm-images'
 env:
   CIRRUS_LOG_TIMESTAMP: true
 
+  # Prefix for image names
+  PREFIX: pg-ci
+
   # See gcp_project_setup.txt for setup instructions
   GCP_PROJECT_MAIN: pg-ci-images
   GCP_PROJECT_DEV: pg-ci-images-dev
@@ -80,7 +83,8 @@ gcp_docker_auth_win: &gcp_docker_auth_win
 
 
 task:
-  name: cleanup-leftover-running
+  name: Clean up leftover running
+  alias: cleanup-leftover-running
 
   container:
     dockerfile: docker/linux_debian_packer
@@ -94,9 +98,11 @@ task:
 
 
 task:
+  name: 'Building VM Image: ${IMAGE_NAME}'
+  alias: vmbuild-${TASK_NAME}
   matrix:
-    - name: freebsd-13
-      env:
+    - env:
+        TASK_NAME: freebsd-13
         PACKERFILE: packer/freebsd.pkr.hcl
 
     - env:
@@ -104,18 +110,27 @@ task:
         SCRIPTS: scripts/linux_debian_*
 
       matrix:
-        - name: bullseye
-        - name: sid
-        - name: sid-newkernel
-        - name: sid-newkernel-uring
+        - env:
+            TASK_NAME: bullseye
+        - env:
+            TASK_NAME: sid
+        - env:
+            TASK_NAME: sid-newkernel
+        - env:
+            TASK_NAME: sid-newkernel-uring
 
     - env:
         PACKERFILE: packer/windows.pkr.hcl
         SCRIPTS: scripts/windows*
 
       matrix:
-        - name: windows-ci-vs-2019
-        - name: windows-ci-mingw64
+        - env:
+            TASK_NAME: windows-ci-vs-2019
+        - env:
+            TASK_NAME: windows-ci-mingw64
+
+  env:
+    IMAGE_NAME: ${PREFIX}-${TASK_NAME}
 
   depends_on: cleanup-leftover-running
 
@@ -139,11 +154,11 @@ task:
 
   build_image_script: |
     packer build \
-      -only "*.${CIRRUS_TASK_NAME}" \
+      -only "*.${TASK_NAME}" \
       -timestamp-ui \
       -force \
       -var "image_date=$DATE" \
-      -var "task_name=${CIRRUS_TASK_NAME}" \
+      -var "image_name=${IMAGE_NAME}" \
       -var gcp_project=$GCP_PROJECT \
       "$PACKERFILE"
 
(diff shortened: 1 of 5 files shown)


.cirrus.yml
```diff
@@ -9,6 +9,22 @@ env:
...
+  SID_NEWKERNEL_URING: sid-newkernel-uring
+  WINDOWS_CI_VS_2019_VM: windows-ci-vs-2019
+  WINDOWS_CI_VS_MINGW64_VM: windows-ci-mingw64
+  OPENBSD_VANILLA: openbsd-vanilla
+  OPENBSD_POSTGRES: openbsd-postgres
+  NETBSD_VANILLA: netbsd-vanilla
+  NETBSD_POSTGRES: netbsd-postgres
+  LINUX_DEBIAN_BULLSEYE_CI: linux_debian_bullseye_ci
+  WINDOWS_CI_VS_2019_CONTAINER: windows_ci_vs_2019
+  WINDOWS_CI_VS_MINGW64_CONTAINER: windows_ci_mingw64
```
anarazel, line 26: I don't really see the point of these variables? They're longer than the names they end up as?

.cirrus.yml
```diff
@@ -205,15 +241,15 @@ task:
...
-    - windows-ci-mingw64
+    - 'Building VM Image: ${FREEBSD_13}'
+    - 'Building VM Image: ${NETBSD_POSTGRES}'
+    - 'Building VM Image: ${OPENBSD_POSTGRES}'
+    - 'Building VM Image: ${BULLSEYE}'
+    - 'Building VM Image: ${SID}'
+    - 'Building VM Image: ${SID_NEWKERNEL}'
+    - 'Building VM Image: ${SID_NEWKERNEL_URING}'
+    - 'Building VM Image: ${WINDOWS_CI_VS_2019_VM}'
+    - 'Building VM Image: ${WINDOWS_CI_VS_MINGW64_VM}'
```
anarazel, line 252: We could define an alias for the tasks to make this a bit less annoying. See:
https://cirrus-ci.org/guide/writing-tasks/#task-execution-dependencies

.cirrus.yml
nbyavuz: I wanted to use these variables for both tasks and the depends_on. When there is a change on the task name, same change should be applied to depends_on. These variables could solve this problem but like you said alias could work, too.

anarazel: But in that case the variable name should be changed as well.  So the indirection doesn't buy us anything.

nbyavuz:
Since we didn't talk about that, I added 'refactoring remaining tasks' as a second commit.

.cirrus.yml
```diff
@@ -217,7 +239,7 @@ task:
 # Only do so after a run successfully creating all images, to avoid deleting
 # all image versions when image creation has failed for a few days.
 task:
-  name: cleanup-old
+  name: Clean up old VM images
```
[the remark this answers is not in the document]
anarazel: Ping?

.cirrus.yml
```diff
@@ -420,15 +448,15 @@ task:
...
+    - name: 'Testing VM Image: windows-ci-mingw64'
       depends_on:
         - windows-ci-mingw64
       compute_engine_instance:
         image_project: $GCP_PROJECT
         image: family/pg-ci-windows-ci-mingw64
         platform: windows
 
-    - name: test-windows-mingw64-amd64-container
+    - name: 'Testing Docker Image: windows_ci_mingw64'
```
anarazel, line 459: I think s/Docker Image/Container Image/ would be good, here and in earlier locations.

.cirrus.yml
```diff
@@ -89,28 +89,42 @@ on_failure_googlecompute: &on_failure_googlecompute
 
 
 task:
+  name: 'Building VM Image: ${IMAGE_NAME}'
+  alias: ${IMAGE_NAME}
```
anarazel, line 93: It might be good if the task aliases had vmbuild- or such as a prefix, otherwise we'll just end up with conflicts vs containers?

.cirrus.yml
```diff
@@ -89,28 +89,42 @@ on_failure_googlecompute: &on_failure_googlecompute
...
 
-      matrix:
-        - name: bullseye
-        - name: sid
-        - name: sid-newkernel
-        - name: sid-newkernel-uring
+    - env:
+        IMAGE_NAME: sid-newkernel-uring
+        PACKERFILE: packer/linux_debian.pkr.hcl
+        SCRIPTS: scripts/linux_debian_*
```
anarazel, line 117: Hm, what's the advantage of getting rid of this matrix?

.cirrus.yml
```diff
@@ -344,9 +373,9 @@ docker_builder:
 
 
 task:
-  name: test-linux-container-amd64
+  name: 'Testing Docker Image: linux_debian_bullseye_ci'
```
anarazel, line 376: Why do we need the image name in two places here?

.cirrus.yml
```diff
@@ -131,42 +145,50 @@ task:
...
-      -var "task_name=${CIRRUS_TASK_NAME}" \
+      -var "task_name=${IMAGE_NAME}" \
       -var gcp_project=$GCP_PROJECT \
       "$PACKERFILE"
     gcloud compute images add-iam-policy-binding \
       --project "$GCP_PROJECT" \
       --member=allAuthenticatedUsers \
       --role=roles/compute.imageUser \
-      pg-ci-${CIRRUS_TASK_NAME}-${DATE}
+      pg-ci-${IMAGE_NAME}-${DATE}
```
anarazel, line 159: It feels wrong that some of this references pg-ci-${IMAGE_NAME}, other parts just ${IMAGE_NAME}.

.cirrus.yml
```diff
@@ -131,42 +145,50 @@ task:
...
-        - name: openbsd-postgres
-          depends_on:
-            - openbsd-vanilla
 
     - env:
-        PACKERFILE: packer/netbsd_openbsd.pkr.hcl
+        IMAGE_NAME: openbsd-postgres
+        PKRVARFILE: packer/openbsd.pkrvars.hcl
+      depends_on:
+        - openbsd-vanilla
```
anarazel, line 178: Similar to above: Why are we getting rid of the matrix?

anarazel:
@nbyavuz Could you rebase?

.cirrus.yml
nbyavuz: There are dash differences, if we add vmbuild or such we should set depends_on for each test task. Does it worth it?

For example, current task:
```yaml
  matrix:
    - name: 'Testing VM Image: ${IMAGE_NAME}'
      env:
        IMAGE_NAME: windows-ci-mingw64
      compute_engine_instance:
        image_project: $GCP_PROJECT
        image: family/pg-ci-${IMAGE_NAME}
        platform: windows

    - name: 'Testing Container Image: ${IMAGE_NAME}'
      env:
        IMAGE_NAME: windows_ci_mingw64
      windows_container:
        image: $GCP_REPO/${IMAGE_NAME}:latest

  depends_on:
    - ${IMAGE_NAME}
```

.cirrus.yml
nbyavuz: Yes, that bothers me too. Maybe we can set global prefix environment variable and remove prefix from packer files? Then, we can set IMAGE_NAME something like ${PREFIX}-image_name.

.cirrus.yml
nbyavuz: Fixed.

.cirrus.yml
anarazel: I like it - eventually we could use that to name images differently for PRs for example.

.cirrus.yml
anarazel: > There are dash differences, if we add vmbuild or such we should set depends_on for each test task. Does it worth it?

That's not that big a difference to today, is it? You can still continue to use ${IMAGE_NAME}.

.cirrus.yml
```diff
@@ -246,16 +266,17 @@ task:
   # deal with multi-arch containers (gcloud shows the individual architecture
   # specific containers without tags, even if the manifest has tags)
   cleanup_container_images_script: |
-    gcloud artifacts docker images list \
+    gcloud artifacts Container Images list \
```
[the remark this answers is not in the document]
anarazel: Hm, you resolved this, but afaics the code looks the same, and would still not work?

packer/linux_debian.pkr.hcl
```diff
@@ -1,34 +1,28 @@
...
     },
     {
-      name = "sid-newkernel"
+      source_name = "sid-newkernel"
       zone = "us-west2-a"
       machine = "c2-standard-8"
     },
     {
-      name = "sid-newkernel-uring"
+      source_name = "sid-newkernel-uring"
```
anarazel, line 25: What does "source_name" really mean? With "source" I'd imagine it to reference the original image or such, but clearly this isn't that.

nbyavuz: It is packer's build's source_name; I found the 'name' itself confusing so I changed it to source_name. Does 'build_source_name/id' or such work?

anarazel: I don't know what "packer's build's source_name" means?

nbyavuz: [This source part](https://github.com/anarazel/pg-vm-images/blob/35e3a3e9fd1ffd225059d820c63924aaa2671c72/packer/linux_debian.pkr.hcl#LL53C1-L65C4)

I mean packer files have source part and we set this source part's name using dictionaries(`debian_gcp_images` on this example). We set these sources to match with `packer build -only "*.${TASK_NAME}" `.

I think it being only stated as `name` is confusing since we also have `IMAGE_NAME` variable and this `name` is not same as `IMAGE_NAME`. So, I wanted to change it  to `source_name` since it is only used while we dynamically set source part.

nbyavuz: I think they need to be same with `TASK_NAME`, maybe we can set these variables as `task_name` instead of `source_name`?

anarazel:
There are conflicts, could you rebase?

anarazel (approved):
We could do more here, but I think it's good enough!

PR #58, opened 2023-03-27
Use same date for all images

nbyavuz:
There is a CIRRUS_CHANGE_TIMESTAMP variable which is timestamp of when the commit is pushed. Convert that timestamp to date and use it at the images. So, all of the images' date part will be the same. Then we can use and update these images at the other CI tasks.

A downside comes to my mind, if the image is created and we re-run CI; it will fail because the image name will be same.


anarazel:
I don't think that works - the images are rebuilt daily, and that will all use the same commit, no?

nbyavuz:
Actually I thought the same but tested and found that CIRRUS_CHANGE_TIMESTAMP is different for cron jobs. For example:

These four tasks have same commit hash: d1ef15b8632ebc99115a29f232e2c7e20a45e075
but have different timestamps. All timestamps show the timestamp of the when cron job started.

https://cirrus-ci.com/task/6486504743108608 -> CIRRUS_CHANGE_TIMESTAMP=1679607024158
https://cirrus-ci.com/task/4801905692508160 -> CIRRUS_CHANGE_TIMESTAMP=1679520624502
https://cirrus-ci.com/task/5849622967484416 -> CIRRUS_CHANGE_TIMESTAMP=1679434223266
https://cirrus-ci.com/task/5030596091576320 -> CIRRUS_CHANGE_TIMESTAMP=1679347823902

anarazel:
Oh, interesting.

.cirrus.yml
```diff
@@ -17,6 +17,15 @@ env:
   # GCP_REGION
   # GCP_REPO
 
+# Get the same date for the tasks. CIRRUS_CHANGE_TIMESTAMP is
+# timestamp of when the commit is pushed in ms. Add 500ms to
```
anarazel, line 21: Should note that the timestamp does change when task is triggered by cron.

.cirrus.yml
```diff
@@ -17,6 +17,15 @@ env:
...
+# round up to the nearest second then convert it to date. Another method
+# is used at the Windows docker-builder task because $CIRRUS_ENV can't
+# be updated and this method is not working on PowerShell.
```
anarazel, line 24: Why can't CIRRUS_ENV be updated? IIRC it was broken at some point, but after I reported that, cirrus implemented a fix?

nbyavuz: I couldn't update them on Windows, it says: 
'Failed collect CIRRUS_ENV subsystem results: CIRRUS_ENV file should contain lines in KEY=VALUE format'
but it is already in KEY=VALUE format.

[Example task](https://cirrus-ci.com/task/5672164246945792)

nbyavuz:
> What about this alternative: We have a first task that does nothing but export the date into a text artifcact, that the other tasks then use to determine the timestamp?

I didn't understand this, doesn't that text file stay on first task's machine, how will the other tasks reach it?

nbyavuz:
> Hm. I wonder if it's going to be confusing that an image with newer packages could have an older timestamp?

This should happen rarely, right? Otherwise, yes. It can be confusing.

.cirrus.yml
anarazel: I suspect that's an encoding issue. Powershell will write utf-16 by default.

anarazel: Ping ^

anarazel (commented):
I wonder if a simpler solution could be implemented with starlark:
https://cirrus-ci.org/guide/programming-tasks/

We could make the starlark script emit something like
```yaml
env:
   IMAGE_DATE: iso-date-string
```

which I think should suffice?  We already do that for some vars, so it should be easy:
https://github.com/anarazel/pg-vm-images/blob/main/.cirrus.star

Apparently this should give us the time:
https://github.com/qri-io/starlib/blob/master/time/doc.go#LL27C1-L27C1

nbyavuz:
> Apparently this should give us the time: https://github.com/qri-io/starlib/blob/master/time/doc.go#LL27C1-L27C1

It seems [time](https://github.com/cirruslabs/cirrus-ci-docs/discussions/1173#discussioncomment-6100946) is not available in Cirrus Starlark.

PR #62, opened 2023-05-22
ci: Delete VMs older than 23 hours

nbyavuz:
There could be some leftover VMs. Clear these VMs because of their costs and to get back the space their disks occupy.

It is set to get VMs older than 23 hours instead of 1 day because cron job runs daily, so there could be race conditions.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -80,6 +80,20 @@ gcp_docker_auth_win: &gcp_docker_auth_win
 
 
 task:
+  name: cleanup-leftover-running
+
+  container:
+    dockerfile: docker/linux_debian_packer
+    cpu: 0.5
+    memory: 256Mi
+
+  <<: *gcp_auth_unix
+
+  cleanup_leftover_running_script:
+    ./gcp_delete_leftover_running.py
+
+
+task:
   matrix:
     - name: freebsd-13
       env:
@@ -102,6 +116,8 @@ task:
       matrix:
         - name: windows-ci-vs-2019
         - name: windows-ci-mingw64
+
+  depends_on: cleanup-leftover-running
 
   container:
     dockerfile: docker/linux_debian_packer
@@ -151,6 +167,8 @@ task:
         VERSION: 7-2
       matrix:
         - name: openbsd-vanilla
+          depends_on:
+            - cleanup-leftover-running
         - name: openbsd-postgres
           depends_on:
             - openbsd-vanilla
@@ -161,6 +179,8 @@ task:
         VERSION: 9-3
       matrix:
         - name: netbsd-vanilla
+          depends_on:
+            - cleanup-leftover-running
         - name: netbsd-postgres
           depends_on:
             - netbsd-vanilla
@@ -314,6 +334,8 @@ docker_builder:
 
   platform: windows
   os_version: 2019
+
+  depends_on: cleanup-leftover-running
 
   skip: $CIRRUS_LAST_GREEN_CHANGE != '' && $CIRRUS_CRON != 'regular-rebuild' && !changesInclude('.cirrus.yml', $PACKERFILE, 'scripts/windows*')
 
--- a/gcp_delete_leftover_running.py
+++ b/gcp_delete_leftover_running.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+
+# Delete leftover runnings older than 23 hour
+
+import os
+import subprocess
+import sys
+
+
+def delete_leftover_running(type: str):
+    base_cmd = ['gcloud', 'compute', '--project',
+                os.environ['GCP_PROJECT'], type]
+
+    # get objects older than 23 hours with their names and zones
+    # 23 hour is used instead of 1 day because cron job runs daily,
+    # so there could be race conditions
+    get_cmd = base_cmd + ['list', '--format', 'object value(name, zone)',
+                          '--filter', 'creationTimestamp < -P23H']
+    res = subprocess.run(get_cmd, capture_output=True,
+                         check=True, text=True)
+
+    delete_list = []
+    for line in res.stdout.strip().split('\n'):
+        # check if line is not empty
+        if line:
+            linesplit = line.split()
+            delete_list.append({'name': linesplit[0],
+                               'zone': linesplit[1]})
+
+    if delete_list:
+        print(f"{type} to be deleted: ", ', '.join(
+            disk['name'] for disk in delete_list))
+    else:
+        print(f"no {type} to delete")
+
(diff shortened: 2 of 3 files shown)


gcp_delete_old_images_and_vms.py
```diff
@@ -0,0 +1,102 @@
...
+    if delete_vms:
+        print("VMs to be deleted: ", ', '.join(
+            vm['name']for vm in delete_vms))
+
+    # delete old VMs
+    for vm in delete_vms:
+        print('Deleting VM =', vm['name'])
+        delete_cmd = base_cmd + ['delete', vm['name'],
+                                 '--zone', vm['zone'], '--quiet']
+        subprocess.run(delete_cmd, check=True)
```
anarazel, line 89: I think we also need to delete old disks for those VMs?

gcp_delete_old_images_and_vms.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def main():
+    if 'GCP_PROJECT' not in os.environ:
+        print("GCP_PROJECT not set", file=sys.stderr)
+        sys.exit(1)
+
+    delete_vms()
+    # delete_families()
```
anarazel, line 98: what's that about?

gcp_delete_old_images_and_vms.py
nbyavuz: This command actually deletes VMs and their disks.

gcp_delete_old_images_and_vms.py
nbyavuz: My bad, sorry.

gcp_delete_old_images_and_vms.py
anarazel: I think that's only true for disks explicitly owned by VMs - but the way packer creates disk they are not set up that way (because packer needs to access it after shutting the VM down).

nbyavuz: Added a explicit way to deleting remaining disks older than 23 hours

gcp_delete_old_images_and_instances.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
```
anarazel, line 1: Could you put the rename in a separate commit? Unfortunately it makes the diff hard to read...

gcp_delete_old_images_and_instances.py
```diff
@@ -0,0 +1,104 @@
...
+
+import os
+import subprocess
+import sys
+
+
+# Delete images older than 2 weeks, except if an image is the newest image in
+# an image family
+def delete_families():
```
anarazel, line 10: This does *not* delete families. It deletes *older* images in families, but never the last image in a family.

gcp_delete_old_images_and_instances.py
```diff
@@ -0,0 +1,104 @@
...
+
+    print("deleting images: ", ', '.join(delete_images))
+
+    # finally delete old images
+    delete_cmd = base_cmd + ['delete', '--quiet'] + delete_images
+    subprocess.run(delete_cmd, check=True)
+
+
+# Delete objects older than 23 hour
+def delete_objects(type: str):
```
anarazel, line 60: I don't think "objects" is a good name - that'd include images, which are handled separately.  Maybe delete_leftover_running()?

gcp_delete_old_images_and_instances.py
```diff
@@ -0,0 +1,104 @@
...
+        subprocess.run(delete_cmd, check=True)
+
+
+def main():
+    if 'GCP_PROJECT' not in os.environ:
+        print("GCP_PROJECT not set", file=sys.stderr)
+        sys.exit(1)
+
+    delete_objects('instances')
+    delete_objects('disks')
```
anarazel, line 99: I think we should delete the instances / disks in a separate script. We only run the cleanup task if everything succeeded, but we are more likely to accumulate leftover things if there are failures. 

Also, if there are leftover objects, we might fail running due to resource constraints. So we should remove leftover things *before* we run packer.

nbyavuz:
I didn't revert changes on 'gcp_delete_old_images.py' file, they are mostly styling changes(using pycodestyle). Should I revert these changes too?

.cirrus.yml
```diff
@@ -151,6 +167,8 @@ task:
         VERSION: 7-2
       matrix:
         - name: openbsd-vanilla
+          depends_on:
+           - cleanup-leftover-running
```
anarazel, line 171: Can't we do this depend outside of the matrix, instead of repeating it for each "arm"?

gcp_delete_leftover_running.py
```diff
@@ -0,0 +1,52 @@
...
+                               'zone': linesplit[1]})
+
+    if delete_list:
+        print(f"{type} to be deleted: ", ', '.join(
+            disk['name'] for disk in delete_list))
+    else:
+        print(f"no {type} to delete")
+
+    for elem in delete_list:
+        print(f'Deleting {type} =', elem['name'])
```
anarazel, line 36: If you're using f-strings, why not also use them for `elem['name']`?

anarazel (commented):
Could you split the change to delete-old-images into a separate PR or at least separate commit?

.cirrus.yml
nbyavuz: If it is implemented like you said, 'depends_on at outside of matrix' overwrites open/netbsd-postgres task's depends_on. They starts to depend 'cleanup-leftover-running' instead of their vanilla tasks.

anarazel (approved):
Looks good to me. Have you been able to test that this actually works as-is?

nbyavuz:
> Looks good to me. Have you been able to test that this actually works as-is?

Yes, I set time to 1 second and run it locally:

```sh
$ python3 gcp_delete_leftover_running.py 
instances to be deleted:  instance-1, instance-2
Deleting instances = $instance-1
Deleted [https://www.googleapis.com/compute/v1/projects/pg-ci-images-dev/zones/us-central1-a/instances/instance-1].
Deleting instances = $instance-2
Deleted [https://www.googleapis.com/compute/v1/projects/pg-ci-images-dev/zones/us-central1-a/instances/instance-2].
disks to be deleted:  disk-1
Deleting disks = $disk-1
Deleted [https://www.googleapis.com/compute/v1/projects/pg-ci-images-dev/zones/us-central1-a/disks/disk-1].
```

Just pushed a small fix for removing '$' sign while printing.

anarazel:
> Yes, I set time to 1 second and run it locally:

Cool, then let's merge this once the tests have passed.

PR #72, opened 2023-06-23
ci: Use c2-standard-4 instances

nbyavuz:
Use c2-standard-4 for the all instances since it is cheaper compared to other instance types.

### DIFF
--- a/packer/linux_debian.pkr.hcl
+++ b/packer/linux_debian.pkr.hcl
@@ -15,12 +15,12 @@ locals {
     {
       name = "bullseye"
       zone = "us-west1-a"
-      machine = "e2-highcpu-4"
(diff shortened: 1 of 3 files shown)


packer/linux_debian.pkr.hcl
```diff
@@ -15,22 +15,18 @@ locals {
...
     {
       name = "sid-newkernel"
       zone = "us-west2-a"
-      machine = "c2-standard-8"
     },
     {
       name = "sid-newkernel-uring"
       zone = "us-west2-a"
-      machine = "c2-standard-8"
     },
```
anarazel, line 30: Note that these, intentionally, were using c2-standard-8 (i.e more cores).

anarazel (approved):

PR #75, opened 2023-07-13
ci: Delete old docker images

nbyavuz:
When it tries to delete old docker images, it fails with `Precondition check failed` error. This could be related to permission issues. I am creating this PR as a draft because I need change `createTime filter` to 2 weeks and test if it is a permission issue.

### DIFF
--- a/gcp_delete_old_images.py
+++ b/gcp_delete_old_images.py
@@ -3,12 +3,15 @@
 # Delete images older than 2 weeks, except if an image is the newest image in
 # an image family
 
+import json
 import os
 import subprocess
 import sys
 
 
-def delete_old_images():
+def delete_old_vm_images():
+    print("Deleting VM images")
+
     base_cmd = ['gcloud', 'compute', '--project',
                 os.environ['GCP_PROJECT'], 'images']
 
@@ -47,8 +50,8 @@ def delete_old_images():
             delete_images.append(old_image)
 
     if len(delete_images) == 0:
-        print("no images to delete")
-        sys.exit(0)
+        print("no VM images to delete")
+        return
 
     print("deleting images: ", ', '.join(delete_images))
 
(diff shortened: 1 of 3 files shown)


nbyavuz:
They failed with same error: 

https://cirrus-ci.com/task/6743658368598016?logs=cleanup_vm_images#L24

```
[14:05:11.645] Digests:
[14:05:11.645] - us-docker.pkg.dev/pg-ci-images-dev/ci/linux_debian_bookworm_ci@sha256:0f74711de4e946afbe0aeef1b103b7e8a115c004b827b454d9e5a57b1c043fea
[14:05:11.726] Delete request issued.
[14:05:11.791] Waiting for operation [projects/pg-ci-images-dev/locations/us/operations/981c6dc0-bc62-413a-bc60-063eceaf0836] to complete...
[14:05:12.845] .....failed.
[14:05:12.998] ERROR: (gcloud.artifacts.docker.images.delete) Precondition check failed.
[14:05:14.804] Digests:
[14:05:14.804] - us-docker.pkg.dev/pg-ci-images-dev/ci/linux_debian_bookworm_ci@sha256:376b0e1ff1e4b302289c4ad11f8ceaef39c993568ff181426848e16445ca7235
[14:05:14.865] Delete request issued.
[14:05:14.951] Waiting for operation [projects/pg-ci-images-dev/locations/us/operations/4bab4986-8146-4a04-be42-fbd04c9f6b1a] to complete...
[14:05:16.010] .....failed.
[14:05:16.170] ERROR: (gcloud.artifacts.docker.images.delete) Precondition check failed.
[14:05:17.938] Digests:
[14:05:17.938] - us-docker.pkg.dev/pg-ci-images-dev/ci/linux_debian_bookworm_ci@sha256:5f683ed1f2b6ee170fed06e1a6cd6929d5e2dda583b9eb2fb0c69396da8b7eed
[14:05:18.004] Delete request issued.
[14:05:18.041] Waiting for operation [projects/pg-ci-images-dev/locations/us/operations/9124f650-4393-4ac3-85c2-20ccef778541] to complete...
[14:05:19.083] .....failed.
[14:05:19.248] ERROR: (gcloud.artifacts.docker.images.delete) Precondition check failed.
[14:05:21.025] Digests:
[14:05:21.025] - us-docker.pkg.dev/pg-ci-images-dev/ci/linux_debian_bookworm_ci@sha256:9cfea0875ef3ea686368eaf577b2c8ee2a95794592638e66e550af711868ffe9
[14:05:21.098] Delete request issued.
[14:05:21.137] Waiting for operation [projects/pg-ci-images-dev/locations/us/operations/4e353b46-4465-41ba-ad9d-2193e33e2e49] to complete...
[14:05:22.191] .....failed.
[14:05:22.347] ERROR: (gcloud.artifacts.docker.images.delete) Precondition check failed.
```

Google says: 
https://cloud.google.com/sdk/gcloud/reference/artifacts/docker/images/delete

```
 This command can fail for the following reasons:

*    Trying to delete an image by digest when the image is still tagged. Add --delete-tags to delete the digest and the tags.
*    Trying to delete an image by tag when the image has other tags. Add --delete-tags to delete all tags.
*    A valid repository format was not provided.
*    The specified image does not exist.
*    The active account does not have permission to delete images.
```

Could the reason be permission issues because others doesn't make much sense?

anarazel:
I think the images that fail to be deleted are somehow part of the other images. I was unable to interactively delete those versions as well, but once I deleted one of the tagged images, I could also delete the other images created at the same time.

anarazel:
I think what might work is to first delete all images that are tagged and old enough, and then in a subsequent step, delete all images that are untagged and old enough?  I'd just one search, and then manually implement that on the client side, since doing it with --filter seems to require deprecated matching syntax (or I am too stupid to figure it out).

anarazel:
Ah, figured out how to see this stuff:
```
podman manifest inspect us-docker.pkg.dev/pg-ci-images-dev/ci/linux_debian_bookworm_ci:latest
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 864,
            "digest": "sha256:5f683ed1f2b6ee170fed06e1a6cd6929d5e2dda583b9eb2fb0c69396da8b7eed",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 566,
            "digest": "sha256:0f74711de4e946afbe0aeef1b103b7e8a115c004b827b454d9e5a57b1c043fea",
            "platform": {
                "architecture": "unknown",
                "os": "unknown"
            }
        }
    ]
}
```
Those SHAs match the images that couldn't be deleted. So it's really just that they're referenced from the tagged image. Extremely unhelpful representation.

nbyavuz:
> I think what might work is to first delete all images that are tagged and old enough, and then in a subsequent step, delete all images that are untagged and old enough? I'd just one search, and then manually implement that on the client side, since doing it with --filter seems to require deprecated matching syntax (or I am too stupid to figure it out).

I tested and confirm that is working. Only one minor point is that the script filters latest image but its manifests are not tagged as latest. So, script will try to remove these manifests but it will get 'Precondition check failed.' error since the actual image is not deleted. This can happen rarely since there is a 14 day limitation. Because of this , I didn't add `check=True` while running dockers delete_cmd command but since its rare we can ignore it and add `check=True`.

gcp_delete_old_images.py
```diff
@@ -57,12 +59,59 @@ def delete_old_images():
...
+
+    if len(old_images) == 0:
+        print("no docker images to delete")
+        sys.exit(0)
+
+    # latest image has two tags(date and latest tag), so it will not
+    # be deleted since its len will be 4
+    images_with_tags = [image.split() for image in old_images
+                        if len(image.split()) == 3]
+    images_without_tags = [image.split() for image in old_images
```
anarazel, line 101: This seems pretty ugly. Why not check if the tag actually contains latest?

gcp_delete_old_images.py
```diff
@@ -57,12 +59,59 @@ def delete_old_images():
...
+
+
+def delete_old_docker_images():
+    print('\nDeleting docker images')
+    base_cmd = ['gcloud', 'artifacts', 'docker', 'images']
+    get_images_cmd = base_cmd + ['list',
+                                 '--include-tags', '--format',
+                                 'value[separator=" "](package,version,tags)',
+                                 '--filter', 'createTime < -P2W',
+                                 os.environ['GCP_REPO']]
```
anarazel, line 87: I suggest using --output json or something like that, so we have a useful output format to work with.

gcp_delete_old_images.py
```diff
@@ -57,12 +59,59 @@ def delete_old_images():
...
+
+    delete_images = [f'{image_name[0]}@{image_name[1]}' for
+                     image_name in delete_images]
+
+    print(f"deleting docker images {'with tags' if with_tags else 'without tags'}:\n" +
+          '\n'.join(delete_images))
+
+    for image in delete_images:
+        # '--delete-tags' flag shouldn't be added when deleting images without
+        # tags to catch unseen errors
```
anarazel, line 74: So why not remove the flag then?

nbyavuz: I didn't understand this. Images with tags need this flag but others don't.

nbyavuz:
> It's not great. We could do better, by using podman / docker to inspect the manifests etc. But I doubt it's worth it for the moment.

I used podman like you said and filtered images. It doesn't try to delete the latest images' manifests now.
Example task: https://cirrus-ci.com/task/6364608982679552

gcp_delete_old_images.py
anarazel: "shouldn't be added" sounds like you're adding it, but it shouldn't be.

gcp_delete_old_images.py
```diff
@@ -57,12 +60,97 @@ def delete_old_images():
...
+                                 'json(package,version,tags)',
+                                 '--filter', 'createTime < -P2W',
+                                 os.environ['GCP_REPO']]
+    res = subprocess.run(get_images_cmd, capture_output=True,
+                         check=True, text=True)
+    old_images = json.loads(res.stdout)
+
+    if len(old_images) == 0:
+        print("no docker images to delete")
+        sys.exit(0)
```
anarazel, line 114: I don't think this should use sys.exit(), given that we do multiple things in the script now. I'd just return.

gcp_delete_old_images.py
```diff
@@ -57,12 +60,97 @@ def delete_old_images():
...
+            latest_images.append(image)
+        elif image['tags']:
+            images_with_tags.append(image)
+        else:
+            images_without_tags.append(image)
+
+    # find the latest images' manifests
+    latest_images_manifests = get_manifests(latest_images)
+
+    # # filter to-be-deleted images by the latest images' manifests
```
anarazel, line 130: Double comment

gcp_delete_old_images.py
```diff
@@ -57,12 +60,97 @@ def delete_old_images():
...
+                  "it is manifest of one of the latest images'")
+        else:
+            images_without_tags_filtered.append(image)
+
+    # delete these images
+    delete_old_docker_images_helper(images_with_tags, base_cmd,
+                                    latest_images_manifests,
+                                    with_tags=True)
+    delete_old_docker_images_helper(images_without_tags_filtered, base_cmd,
+                                    latest_images_manifests)
```
anarazel, line 144: Can you add a comment explaining why we do this twice?

anarazel (approved):

PR #78, opened 2023-07-13
Bump OpenBSD version in Cirrus task too.

xiaq:
The actual version was bumped in https://github.com/anarazel/pg-vm-images/pull/60, but the version in .cirrus.yml wasn't bumped, meaning the image family still ended with 7-2.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -165,7 +165,7 @@ task:
     - env:
         PACKERFILE: packer/netbsd_openbsd.pkr.hcl
         PKRVARFILE: packer/openbsd.pkrvars.hcl
-        VERSION: 7-2
+        VERSION: 7-3
       matrix:
         - name: openbsd-vanilla
           depends_on:


nbyavuz:
Nice catch, thanks! 

After merging this, we need to bump openBSD version on Postgres tasks too.
I created a PR for that: https://github.com/anarazel/postgres/pull/85

anarazel:
I wonder if we should just remove the version name from at least the postgres image? Regularly having to change the postgres CI task is a PITA.

xiaq:
I'd +1 to that, as a user of those images I also wouldn't have to change the image family in my Cirrus config every time there is a new OpenBSD or NetBSD release.

nbyavuz:
Does it make sense to using only major versions instead of major-minor versions?

anarazel:
Took a bit longer, a bunch of random issues came in the way. 

I'm merging this issue now, to fix the existing inconsistency. But I think it'd be good to rename the images to not have minor versions in there. @xiaq any interest opening a PR?

PR #79, opened 2023-07-17
ci: windows: disable windows container generation

nbyavuz:
Windows container generation will be dropped. Set trigger_type manual for now.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -314,6 +314,9 @@ docker_builder:
     - name: windows_ci_vs_2019
     - name: windows_ci_mingw64
 
+  # disable windows container generation, it will be dropped later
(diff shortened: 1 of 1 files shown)


anarazel:
I tried to figure out whether there's any way to see if there's still uses of the containers.  It doesn't look like there is, completely baffling. 
https://stackoverflow.com/questions/71531832/is-it-possible-track-the-number-of-docker-pulls-in-google-artifact-registry/71532746#71532746
"Unfortunately, the theory about public requests not appearing is correct. CALs are about logging authentication events, and when a request has no authentication whatsover, CALs are not generated."

It sure looks like that hasn't changed....

anarazel (approved):

PR #80, opened 2023-07-17
ci: windows: Merge windows images

### LINKED ISSUE #65: Merge different windows images?
Hi,

Now that we use VMs instead of containers, the image size is not as critical anymore. I think it'd make sense to merge the VS and mingw images. While the image will be larger, the overall storage size will go down. There's a lot of redundant data in the two images, and the image storage costs are starting to be noticable...

Bad idea?

Andres

nbyavuz:
There's a lot of redundant data in the two images, and the image storage
costs are starting to be noticable; so merge these images.

I tested and confirmed it is working. This PR needs to merged together with https://github.com/anarazel/postgres/pull/84.

Closes #65

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -122,12 +122,7 @@ task:
     - env:
         PACKERFILE: packer/windows.pkr.hcl
         SCRIPTS: scripts/windows*
-
-      matrix:
-        - env:
-            TASK_NAME: windows-ci-vs-2019
-        - env:
-            TASK_NAME: windows-ci-mingw64
+        TASK_NAME: windows-ci
 
   env:
     IMAGE_NAME: ${PREFIX}-${TASK_NAME}
@@ -377,7 +372,7 @@ task:
 task:
   name: 'Testing VM Image: ${PREFIX}-${IMAGE_NAME}'
   env:
-    IMAGE_NAME: windows-ci-vs-2019
+    IMAGE_NAME: windows-ci
   depends_on:
     - vmbuild-${IMAGE_NAME}
   compute_engine_instance:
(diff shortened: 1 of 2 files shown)


anarazel:
What are the before and combined images sizes?

nbyavuz:
> What are the before and combined images sizes?

Before:
VS: ~10.5 GB 
Mingw: ~10 GB

Combined: ~11 GB

anarazel:
That's very cool.

.cirrus.yml
```diff
@@ -114,8 +114,7 @@ task:
         SCRIPTS: scripts/windows*
 
       matrix:
-        - name: windows-ci-vs-2019
-        - name: windows-ci-mingw64
+        - name: windows
```
anarazel, line 117: I think it should still have -ci in the name.  We're going to have to update the CI tasks in postgres quickly afterwards...

packer/windows.pkr.hcl
```diff
@@ -245,15 +236,15 @@ build {
         tags = ["${var.image_date}", "latest"]
         # packer version is 1.6.6 while generating the vm images, and it complains
         # if local.only.docker is used here
-        only = ["docker.windows_ci_vs_2019", "docker.windows_ci_mingw64"]
+        only = ["docker.windows_container", "docker.windows_container"]
```
anarazel, line 239: Let's drop the windows containers before we get to this bit - since all users would need to change the name of the container anyway, there's not much point in continuing to build them.

nbyavuz:
Note that this PR is rebased onto https://github.com/anarazel/pg-vm-images/pull/87.

anarazel (approved):
Looks good to me.

PR #85, opened 2023-08-03
Remove version suffix from {Net,Open}BSD image family name.

xiaq:


nbyavuz:
Hi @xiaq, thank you for the PR!

I realized that only minor version needs to be removed, not all version suffix. Could you please update the PR?

anarazel:
> Hi @xiaq, thank you for the PR!
> 
> I realized that only minor version needs to be removed, not all version suffix. Could you please update the PR?

I think it's actually good to not have the version number at all.

anarazel:
Err, that was accidental. I'll open a new PR under "my own name" attributing you as the author.

xiaq:
My $2 is that which version numbers to include should depend on whether you intend to support images across major or minor versions.

OpenBSD has a somewhat interesting versioning scheme - it just goes in increments of 0.1 and it's always the latest two versions that's still supported officially. For OpenBSD the two sensible choices are not having versions at all in the name, or having the full version in the name. Having the major version in the name doesn't make much sense.

NetBSD uses a major.minor versioning scheme, so if you do intend to maintain images for the last two major versions (currently 9.x and 8.x) it would make sense to have the major version in the image name.

xiaq:
> Err, that was accidental. I'll open a new PR under "my own name" attributing you as the author.

No problem at all!

PR #91, opened 2023-09-22
linux: Upgrade Debian from 11 (bullseye) to 12 (bookworm)

nbyavuz:
I tested the new debian 12 images, CI run: https://cirrus-ci.com/build/4851369497067520
```
# install grub again, otherwise grub fails with:
# error: file `/boot/grub/x86_64-efi/bli.mod' not found.
grub-install --target=x86_64-efi --efi-directory=/boot/efi
```

fixes the errors while creating debian 11 images too. I could make it separate commit or create another PR for it.

I didn't update 'linux_debian_packer' to bookworm because 'Clean up old images' started to fail. CI run: https://cirrus-ci.com/task/4831203417653248.
I spent some time but I couldn't fix it yet. The podman command fails with:
- 'overlay is not supported over overlayfs'. I am able to fix it with installing 'fuse-overlayfs'. Then it failed with:
- '/var/lib/containers/storage/overlay flags: 0x1000: permission denied'. I couldn't fix it yet. It is said that it could be related with apparmor on web.

### DIFF
--- a/docker/linux_debian_ci
+++ b/docker/linux_debian_ci
@@ -1,4 +1,6 @@
-FROM debian:bullseye
+ARG DEBIAN_RELEASE
+
+FROM debian:${DEBIAN_RELEASE}
 
 ENV DEBIAN_FRONTEND=noninteractive
 
--- a/packer/linux_debian.pkr.hcl
+++ b/packer/linux_debian.pkr.hcl
@@ -7,36 +7,45 @@ locals {
 
   debian_gcp_images = [
     {
+      task_name = "bookworm"
+      zone = "us-west1-a"
+      machine = "t2d-standard-2"
+      source_image_family = "debian-12"
+    },
+    {
       task_name = "bullseye"
       zone = "us-west1-a"
       machine = "t2d-standard-2"
+      source_image_family = "debian-11"
     },
     {
       task_name = "sid"
       zone = "us-west1-a"
       machine = "t2d-standard-2"
+      source_image_family = "debian-11"
     },
     {
       task_name = "sid-newkernel"
       zone = "us-west1-a"
       machine = "t2d-standard-4"
+      source_image_family = "debian-11"
     },
     {
       task_name = "sid-newkernel-uring"
       zone = "us-west1-a"
       machine = "t2d-standard-4"
+      source_image_family = "debian-11"
     },
   ]
 }
 
-source "googlecompute" "bullseye-vanilla" {
+source "googlecompute" "debian-vanilla" {
   disk_size               = "25"
   disk_type               = "pd-ssd"
   preemptible             = "true"
   project_id              = var.gcp_project
   image_name              = "${local.image_identity}"
   instance_name           = "build-${local.image_identity}"
-  source_image_family     = "debian-11"
   source_image_project_id = ["debian-cloud"]
   ssh_pty                 = "true"
   ssh_username            = "packer"
@@ -53,13 +62,13 @@ build {
 
   dynamic "source" {
     for_each = local.debian_gcp_images
-    labels = ["source.googlecompute.bullseye-vanilla"]
+    labels = ["source.googlecompute.debian-vanilla"]
     iterator = tag
 
     content {
       # can't reference local. / var. here?!?
       name = tag.value.task_name
-
+      source_image_family = tag.value.source_image_family
       zone = tag.value.zone
       machine_type = tag.value.machine
     }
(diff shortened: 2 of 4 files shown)


anarazel:
Why do we need the grub-install bit? I think something is going wrong to leading to needing that. We should work on getting rid of that need, instead of fixing the consequences...

nbyavuz:
> Why do we need the grub-install bit? I think something is going wrong to leading to needing that. We should work on getting rid of that need, instead of fixing the consequences...

It seems removing 'mokutil and shim-unsigned' packages caused that issue but your 1d2ca46ea2 commit already fixed the issue.

packer/linux_debian.pkr.hcl
```diff
@@ -131,6 +131,10 @@ build {
     inline = [
       <<-SCRIPT
         apt-get update -y
+
+        # first update grub2-common, otherwise grub-cloud-amd64 fails
+        DEBIAN_FRONTEND=noninteractive apt-get upgrade grub2-common -y
+
```
anarazel, line 137: Shouldn't be needed anymore, right?

nbyavuz: Unfortunately, no. If I remove this part, CI fails with:

```
linux.googlecompute.sid: Setting up linux-image-6.5.0-2-cloud-amd64 (6.5.6-1) ...
linux.googlecompute.sid: /etc/kernel/postinst.d/initramfs-tools:
linux.googlecompute.sid: update-initramfs: Generating /boot/initrd.img-6.5.0-2-cloud-amd64
linux.googlecompute.sid: /etc/kernel/postinst.d/zz-update-grub:
linux.googlecompute.sid: Generating grub configuration file ...
linux.googlecompute.sid: /etc/grub.d/10_linux: 1: version_find_latest: not found
linux.googlecompute.sid: run-parts: /etc/kernel/postinst.d/zz-update-grub exited with return code 127
linux.googlecompute.sid: dpkg: error processing package linux-image-6.5.0-2-cloud-amd64 (--configure):
linux.googlecompute.sid:  installed linux-image-6.5.0-2-cloud-amd64 package post-installation script subprocess returned error exit status 1
```

packer/linux_debian.pkr.hcl
```diff
@@ -29,14 +29,14 @@ locals {
   ]
 }
 
-source "googlecompute" "bullseye-vanilla" {
+source "googlecompute" "bookworm-vanilla" {
```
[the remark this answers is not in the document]
anarazel: Hm, doesn't look like you updated the branch to do so?

packer/linux_debian.pkr.hcl
anarazel: That's odd. Clearly some broken dependencies...

packer/linux_debian.pkr.hcl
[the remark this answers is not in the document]
anarazel: You did, not sure why I wasn't seeing that yesterday.

packer/linux_debian.pkr.hcl
anarazel: Something seems to have lead to debian sid to fail to build with this PR, even though it's succeeding on main currently:
https://cirrus-ci.com/build/5861999236612096

I suspect it's the change above. Just upgrading grub2-common fails because its dependencies aren't ugraded, and because you're using upgrade, not dist-upgrade, it can't upgrade those, because it'd entail more work than upgrade does.

anarazel: Another thing: Why would this change affect debian sid and thus potentially require adding the above line? Shouldn't the sid build be unaffected?

docker/linux_debian_ci
```diff
@@ -1,6 +1,7 @@
-FROM debian:bullseye
+ARG DATE IMAGE_NAME TASK_NAME
```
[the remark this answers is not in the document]
nbyavuz: Done.

nbyavuz:
I am not sure why but I couldn't reply to your comments about this part:

```
        # first update grub2-common, otherwise grub-cloud-amd64 fails
        DEBIAN_FRONTEND=noninteractive apt-get upgrade grub2-common -y
```

It looks like the problem that required this fix got resolved so I removed this part.

> Another thing: Why would this change affect debian sid and thus potentially require adding the above line? Shouldn't the sid build be unaffected?

Yes, you are right but since this part is removed, I didn't apply this change.

packer/linux_debian.pkr.hcl
```diff
@@ -6,37 +6,46 @@ locals {
...
       task_name = "bullseye"
       zone = "us-west1-a"
       machine = "t2d-standard-2"
+      source_image_family = "debian-11"
     },
     {
       task_name = "sid"
       zone = "us-west1-a"
       machine = "t2d-standard-2"
+      source_image_family = "debian-11"
```
anarazel, line 25: Should probably build sid* based on debian-12...

nbyavuz: I didn't change them because I thought you may not want to upgrade them right now, done.

nbyavuz: When debian-12 is used, all sid* images are failing with:

```
 The following packages have unmet dependencies:
 libkrb5-26t64-heimdal : Breaks: libkrb5-26-heimdal (< 7.8.git20221117.28daf24+dfsg-5)
 E: Unable to correct problems, you have held broken packages.
```

anarazel: I think the problem is just that for bookworm a sources list for bookworm survives in `/etc/apt/sources.list.d/debian.sources`, which leads to the globs matching both bookworm and unstable packages. Leading to the error.

You're right that it might be better to change this separately.

nbyavuz:
Please merge this PR after #95 as both have the same commit.

anarazel (approved):
Looks good. Feel free to merge when the other image build has finished.

PR #92, opened 2024-01-13
Downgrade FreeBSD p5-IO-Tty 1.18->1.17.

macdice:
There seems to be a problem in 1.18 that breaks PostgreSQL's interactive tests.  Until a fixed package rolls out, downgrade to 1.17 if 1.18 is detected.

https://github.com/cpan-authors/IO-Tty/issues/38
https://www.postgresql.org/message-id/flat/757523.1705091568%40sss.pgh.pa.us

### DIFF
--- a/packer/freebsd.pkr.hcl
+++ b/packer/freebsd.pkr.hcl
@@ -93,6 +93,19 @@ build {
           openldap25-client \
           openldap25-server
 
+        # === begin temporary workaround for broken p5-IO-Tty 1.18 ===
+        # https://github.com/cpan-authors/IO-Tty/issues/38
+        # https://www.postgresql.org/message-id/flat/757523.1705091568%40sss.pgh.pa.us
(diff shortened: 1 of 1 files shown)


nbyavuz:
Hi @macdice,

Thank you for the fix!

According to issue you shared, it looks like that problem won't be fixed at least until version 1.20. Do you think adding 1.19 and 1.20 to grep command is a good idea?

macdice:
Not sure, but a couple of people (including me just now) have pinged the maintainer, so I don't think they will be another package released that is broken like this.  So I think it's OK to just look out for the known broken actually released version, and hard to predict the version that will have the fix.

nbyavuz (approved):
The change looks good and I confirm that FreeBSD passes all tests with that change: https://cirrus-ci.com/task/4835937029455872.

macdice:
Thanks!

PR #94, opened 2024-04-24
Fix NetBSD build failures

nbyavuz:
pkgin's repository is not updated when it is installed, it was using NetBSD 9.0's repository, so update pkgin's repository after it is installed.

This should fix recent NetBSD failures.


anarazel:
Thanks!

PR #95, opened 2024-05-02
linux: Move ARG and ENV variables to the end of the dockerfile

nbyavuz:
Using variables whose values could be *different than previous build* causes cache miss. So, move these kind of variables to the end of the dockerfile.

See: https://docs.docker.com/reference/dockerfile/#impact-on-build-caching

Please merge this PR before #91 as both have the same commit.

### DIFF
--- a/docker/linux_debian_bullseye_ci
+++ b/docker/linux_debian_bullseye_ci
@@ -1,7 +1,6 @@
 FROM debian:bullseye
 
-ARG DATE IMAGE_NAME
-ENV DEBIAN_FRONTEND=noninteractive IMAGE_IDENTITY=$IMAGE_NAME-$DATE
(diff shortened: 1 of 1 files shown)


docker/linux_debian_bullseye_ci
```diff
@@ -11,3 +10,7 @@ RUN \
   /scripts/linux_debian_install_deps.sh && \
   rm /scripts/linux_debian_install_deps.sh && \
   apt-get clean && rm -rf /var/lib/apt/lists/*
+
+ARG DATE IMAGE_NAME
```
anarazel, line 14: FWIW I'm not sure it's actually correct to specify multiple ARGs in one line. it's not documented to work.

But that's a separate issue...

anarazel (approved):

PR #97, opened 2024-06-05
freebsd: Update to 13.3

nbyavuz:
freebsd 13.2 is not available on GCP anymore. Also, python2* and python38 packages are removed as well.

PR #98, opened 2024-06-06
Use vcpkg to install Windows dependencies

nbyavuz:
Now `Windows - Server 2019, VS 2019 - Meson & ninja` task able to find `icu, libxml, libxslt, lz4, openssl, plperl, zlib and zstd`. 

There are 4 steps to install packages via vcpkg:

1- Create a base Windows image which has dependencies to run vcpkg.

2- Install packages via vcpkg and save & upload the vcpkg cache. Reuse that cache on the next install.

3- Create a Windows image from the base image (step #1). While installing packages via vcpkg, use cache from step #2.

4- Delete Windows base image.

Example run with vcpkg image: https://cirrus-ci.com/build/4816085688516608

**Important notes:** 

- After this PR is merged, extra_*_dirs needs to be removed from Windows task in Postgres.

- Please merge #97 first.

**Some errors:**

- **tcl** is not working for now and I did not start to debug it yet. I wanted to discuss the PR first since the differences are quite big.

- **readline:** vcpkg is able to install package but build step on Postgres fails.
    ```
    [1527/2241] Compiling C object src/bin/psql/psql.exe.p/tab-complete.c.obj
    FAILED: src/bin/psql/psql.exe.p/tab-complete.c.obj 
    "cl" "-Isrc\bin\psql\psql.exe.p" "-Isrc\bin\psql" "-I..\src\bin\psql" "-Isrc\include" "-I..\src\include" "-I..\src\include\port\win32" "-I..\src\include\port\win32_msvc" "-Isrc/interfaces/libpq" "-I..\src\interfaces\libpq" "-IC:/vcpkg/installed/x64-windows/debug/../include" "/MDd" "/FIpostgres_fe_pch.h" "/Yupostgres_fe_pch.h" "/Fpsrc\bin\psql\psql.exe.p\postgres_fe_pch.pch" "/nologo" "/showIncludes" "/utf-8" "/W2" "/Od" "/Zi" "/DWIN32" "/DWINDOWS" "/D__WINDOWS__" "/D__WIN32__" "/D_CRT_SECURE_NO_DEPRECATE" "/D_CRT_NONSTDC_NO_DEPRECATE" "/wd4018" "/wd4244" "/wd4273" "/wd4101" "/wd4102" "/wd4090" "/wd4267" "/FS" "/FdC:\cirrus\build\src\bin\psql\psql.pdb" /Fosrc/bin/psql/psql.exe.p/tab-complete.c.obj "/c" ../src/bin/psql/tab-complete.c
    ../src/bin/psql/tab-complete.c(4023): fatal error C1026: parser stack overflow, program too complex
    ```
- **libsystemd**: can not be installed via vcpkg.

    ````
    libxcrypt is only supported on 'linux', which does not match x64-windows. This usually means that there are known build failures, or runtime problems, when building other platforms. To ignore this and attempt to build libxcrypt anyway, rerun vcpkg with `--allow-unsupported`.
    ````
- **libuuid**: can not be installed via vcpkg.
    ```
    libuuid is only supported on '!osx & !windows', which does not match x64-windows. This usually means that there are known build failures, or runtime problems, when building other platforms. To ignore this and attempt to build libuuid anyway, rerun vcpkg with `--allow-unsupported`.
    ````

### DIFF
--- a/scripts/windows_install_packages_via_vcpkg.ps1
+++ b/scripts/windows_install_packages_via_vcpkg.ps1
@@ -1,31 +1,112 @@
 # Install packages via vcpkg
+
+param (
+  [switch]$GenerateCacheTask = $false
+)
+
+# param() should be the first thing in the script file,
+# so ErrorActionPreference is defined here
 $ErrorActionPreference = "Stop"
 
-cd ${VCPKG_PATH}
-.\bootstrap-vcpkg.bat -disableMetrics
+Function InstallVcpkgPackages()
+{
+  param (
+    [string]$VCPKG_PATH,
+    [string]$VCPKG_CACHE
+  )
 
-# Build krb5 in optimized mode, it outputs too much information in debug builds
-.\vcpkg.exe install --clean-after-build --binarysource=files,${VCPKG_CACHE},readwrite --triplet=x64-windows `
-  krb5:x64-windows-release
-if (!$?) { throw 'cmdfail' };
+  # If the VCPKG_PATH or VCPKG_CACHE is not provided, try to get it from the environment
+  if (! ($VCPKG_PATH))
+  {
+    $VCPKG_PATH = ${Env:VCPKG_PATH}
+  }
+  if (! ($VCPKG_CACHE))
+  {
+    $VCPKG_CACHE = ${Env:VCPKG_CACHE}
+  }
 
-.\vcpkg.exe install --debug --clean-after-build --binarysource=files,${VCPKG_CACHE},readwrite --triplet=x64-windows `
-  gettext[tools]:x64-windows `
-  icu:x64-windows-static-md `
-  libxml2[tools,iconv,icu]:x64-windows-static-md `
-  libxslt:x64-windows-static-md `
-  lz4:x64-windows-static-md `
-  openssl:x64-windows-static-md `
-  pkgconf:x64-windows-static-md-release `
-  readline-win32:x64-windows `
-  zlib:x64-windows-static-md `
-  zstd:x64-windows-static-md
-if (!$?) { throw 'cmdfail' };
+  cd ${VCPKG_PATH}
+  .\bootstrap-vcpkg.bat -disableMetrics
 
-.\vcpkg.exe export --raw --output=pg-deps --x-all-installed
-if (!$?) { throw 'cmdfail' };
+  # Build krb5 in optimized mode, it outputs too much information in debug builds
+  .\vcpkg.exe install --clean-after-build --binarysource=files,${VCPKG_CACHE},readwrite --triplet=x64-windows `
+    krb5:x64-windows-release
+  if (!$?) { throw 'cmdfail' };
 
-7z.exe a -r pg-deps.7z .\pg-deps\*
-if (!$?) { throw 'cmdfail' };
+  .\vcpkg.exe install --debug --clean-after-build --binarysource=files,${VCPKG_CACHE},readwrite --triplet=x64-windows `
+    gettext[tools]:x64-windows `
+    icu:x64-windows-static-md `
+    libxml2[tools,iconv,icu]:x64-windows-static-md `
+    libxslt:x64-windows-static-md `
+    lz4:x64-windows-static-md `
+    openssl:x64-windows-static-md `
+    pkgconf:x64-windows-static-md-release `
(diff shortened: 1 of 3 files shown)


nbyavuz:
Having the list of the packages to be installed in two places(.cirrus.yml and windows_install_pg_deps.ps1) is bad. I tried to save the list at top env like:

.cirrus.yml:
```yml
env:
  ...
  VCPKG_PACKAGES: gettext icu libxml2 libxslt lz4 openssl pkgconf python3 tcl zlib zstd
```

but powershell gave this error at:
```powershell
.\vcpkg.exe install --debug --binarysource=files,${Env:VCPKG_CACHE},readwrite ${Env:VCPKG_PACKAGES} --triplet=x64-windows
```
and the error message is:
```
error: expected the end of input parsing a package spec; this usually means the indicated character is not allowed to be in a package spec. Port, triplet, and feature names are all lowercase alphanumeric+hypens.
  on expression: gettext icu libxml2 libxslt lz4 openssl pkgconf python3 tcl zlib zstd
```
Do you have any idea why?

macdice:
For readline, there have been a couple of threads about that:

https://www.postgresql.org/message-id/20230607185458.GA1334487%40rfd.leadboat.com

https://www.postgresql.org/message-id/flat/CACLU5mThm-uCtERMVHMoGED2EPfyS54j83WB20s_BmzVQuMkpg%40mail.gmail.com

One of the problems mentioned there is VA_ARGS_NARGS, but we already fixed that in 3f28bd73.

It is interesting that those two threads mention two different workarounds for C1026.  Seems like readline is pretty close to working on Windows, but you should be aware that more work and probably code changes are probably needed!

anarazel:
> Having the list of the packages to be installed in two places(.cirrus.yml and windows_install_pg_deps.ps1) is bad. I tried to save the list at top env like:

Why do we need it in .cirrus.yml at all? Can't we just exclusively have it in some image generation script?

anarazel:
> Do you have any idea why?

I suspect what might be happening is that powershell ends up passing all the packages as a single argument, and that then can't be interpreted as a package name by vcpkg.

nbyavuz:
> There's a lot of stuff in this PR that looks unrelated. Probably needs a rebase?

Rebased. However, I think these unrelated things may be caused by me :worried:.

nbyavuz:
I updated the PR. Now, the PR does not remove anything. It installs packages by vcpkg and to use these packages:

1- Set PKG_CONFIG_PATH to 'c:\vcpkg\installed\x64-windows\debug\lib\pkgconfig'

2- Remove extra_[lib|include]_dirs for the openssl, otherwise there will be a conflict.

If PKG_CONFIG_PATH is not set, it works as it did before.

scripts/windows_install_packages_via_vcpkg.ps1
```diff
@@ -0,0 +1,87 @@
...
+
+  $VCPKG_PKG_PREFIX = "${VCPKG_PATH}\installed\x64-windows"
+
+  $PKG_PATHS = "${VCPKG_PKG_PREFIX}\debug\lib;" +
+    "${VCPKG_PKG_PREFIX}\debug\bin;" +
+    "${VCPKG_PKG_PREFIX}\tools\pkgconf;"
+
+  [Environment]::SetEnvironmentVariable('PATH', ${PKG_PATHS} + [Environment]::GetEnvironmentVariable('PATH', 'Machine'), 'Machine')
+  [Environment]::SetEnvironmentVariable('PKG_CONFIG', 'pkgconf', 'Machine')
+}
```
anarazel, line 73: I wonder if we should set some other environment variable and change the pg test task to set them. That way we're not breaking existing tests.

E.g. something like PG_DEPS_PATH, PG_DEPS_PKG_CONFIG_PATHS? Then it'd just be two short statements to set them in the tests...

nbyavuz:
PR is updated:

1) vcpkg is called once, end image uses result of that.
2) Some packages are installed as static libraries. (copied from Andres' windows-deps-vcpkg-andres branch)
3) krb5 is built in optimized mode.
4) `PG_DEPS_['PATH', 'PKG_CONFIG', 'PKG_CONFIG_PATH', 'EXTRA_INCLUDE_DIRS', 'EXTRA_LIB_DIRS']` are exported as env variables.

PR #103, opened 2024-07-05
bsd: clear_user_and_ssh_keys script should not fail if metadata does not exist

nbyavuz:
clear_user_and_ssh_keys script does not fail when the GCP_PROJECT is 'pg-ci-images-dev' but fails when it is 'pg-ci-images'. Most possible reason is 'pg-ci-images' project does not have at least one of the project_level_ssh_keys and instance_level_ssh_keys. So as a solution, script should not fail if it could not find these keys.

### DIFF
--- a/scripts/bsd/clear_users_and_ssh_keys.sh
+++ b/scripts/bsd/clear_users_and_ssh_keys.sh
@@ -19,9 +19,32 @@ fi
 # these users will exist in postgres images.
 # we need to remove these users.
 
-project_level_ssh_keys=$(curl -fsH "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/project/attributes/ssh-keys)
-instance_level_ssh_keys=$(curl -fsH "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/instance/attributes/ssh-keys)
-ssh_keys=$(printf '%s\n%s' "$project_level_ssh_keys" "$instance_level_ssh_keys")
+ssh_keys=""
+
+# do not fail if curl fails with 404
+curl_helper ()
+{
+    curl_temp_path=$(mktemp)
+
+    exit_code=0
+    http_code=$(curl -fsH "Metadata-Flavor: Google" -o ${curl_temp_path} -w "%{http_code}\n" $1) || exit_code=$?
+
+    if [ "${exit_code}" != "0" ] && [ "${http_code}" != "404" ]
+    then
+        echo "Curl failed with exit code ${exit_code}"
+        exit ${exit_code}
+    elif [ "${http_code}" = "404" ]
+    then
+        echo "Curl returned 404"
+    else
+        ssh_keys=$(printf '%s\n%s' "$ssh_keys" "$(cat $curl_temp_path)")
+    fi
+
+    rm -f $curl_temp_path
+}
+
+curl_helper http://metadata.google.internal/computeMetadata/v1/project/attributes/ssh-keys
+curl_helper http://metadata.google.internal/computeMetadata/v1/instance/attributes/ssh-keys
 
 if [ "$ssh_keys" != "" ]
 then


anarazel:
That seems likely to hide bugs. Could we just ignore 404s or such?

nbyavuz:
> That seems likely to hide bugs. Could we just ignore 404s or such?

I could not find a way to directly check if there is 404, so I used grep for that.

scripts/bsd/clear_users_and_ssh_keys.sh
```diff
@@ -19,9 +19,24 @@ fi
...
-project_level_ssh_keys=$(curl -fsH "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/project/attributes/ssh-keys)
-instance_level_ssh_keys=$(curl -fsH "Metadata-Flavor: Google"  http://metadata.google.internal/computeMetadata/v1/instance/attributes/ssh-keys)
-ssh_keys=$(printf '%s\n%s' "$project_level_ssh_keys" "$instance_level_ssh_keys")
+ssh_keys=""
+
+# do not fail if curl fails with 404
+curl_helper ()
+{
+    exit_code=0
+    curl_output=$(curl -fsSH "Metadata-Flavor: Google" $1 2>&1) || exit_code=$?
```
anarazel, line 28: Why not just do this portion unconditionally? I.e.
```bash
    curl_output=$(curl -fsSH "Metadata-Flavor: Google" $1 2>&1)
    exit_code=$?
```

scripts/bsd/clear_users_and_ssh_keys.sh
```diff
@@ -19,9 +19,24 @@ fi
...
+    if [ "${exit_code}" != 0 ] && (echo ${curl_output} | grep -v "404")
+    then
+        echo "Curl failed with exit code ${exit_code}"
+        exit ${exit_code}
+    else
+        ssh_keys=$(printf '%s\n%s' "$ssh_keys" "$curl_output")
```
anarazel, line 34: Hm, is this actually going to print something sensible in the 404 case, given that you're doing `2>&1`? I'd probably just add a dedicated branch for the 404 situation.

scripts/bsd/clear_users_and_ssh_keys.sh
anarazel: Hm, actually, there might be a better option:
```bash
curl --fail --silent -o /tmp/test -w "%{http_code}\n" 'https://anarazel.de/nofile'
404
```
```bash
curl --fail --silent -o /tmp/test -w "%{http_code}\n" 'https://anarazel.de/'
200
```

nbyavuz: > Hm, actually, there might be a better option:
> 
> ```shell
> curl --fail --silent -o /tmp/test -w "%{http_code}\n" 'https://anarazel.de/nofile'
> 404
> ```
> 
> ```shell
> curl --fail --silent -o /tmp/test -w "%{http_code}\n" 'https://anarazel.de/'
> 200
> ```

That looks cleaner but then curl's output needs to be gotten from `/tmp/test` file and this file needs to be removed as well. Would you prefer that approach?

scripts/bsd/clear_users_and_ssh_keys.sh
nbyavuz: > Hm, is this actually going to print something sensible in the 404 case, given that you're doing `2>&1`? I'd probably just add a dedicated branch for the 404 situation.

I am not sure. -f flag means there will be no body output if curl fails. But I agree with that it is better to add another branch for this case.

scripts/bsd/clear_users_and_ssh_keys.sh
anarazel: Yes - I'd not use /tmp/test though. Use `mktemp` or such.

anarazel (approved):

PR #110, opened 2025-01-13
docker: Install latest packer

nbyavuz:
There were errors while trying to create freeBSD 14.2 image by using the older version of the packer (1.6.6). So, install the latest version.

Latest version of the packer requires its plugins to be installed by using 'pg init' command. This is included in this commit as well.

---

Please note that this is a prerequisite commit for the #109.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -145,6 +145,9 @@ task:
   info_script:
     - packer --version
 
+  packer_init_script:
+    - packer init ${PACKERFILE}
+
   export_date_script: |
     DATE=$(date --utc +'%Y-%m-%dt%H-%M-%S')
     echo "DATE=${DATE}" | tee -a $CIRRUS_ENV
@@ -218,6 +221,9 @@ task:
 
   info_script:
     - packer --version
+
+  packer_init_script:
+    - packer init ${PACKERFILE}
 
   export_date_script: |
     DATE=$(date --utc +'%Y-%m-%dt%H-%M-%S')
--- a/docker/linux_debian_packer
+++ b/docker/linux_debian_packer
@@ -19,9 +19,17 @@ RUN \
   curl \
     -fs \
     -o /usr/share/keyrings/cloud.google.asc \
-    https://packages.cloud.google.com/apt/doc/apt-key.gpg && \
+    https://packages.cloud.google.com/apt/doc/apt-key.gpg \
+    && \
+  echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.asc] https://apt.releases.hashicorp.com bullseye main" > \
+    /etc/apt/sources.list.d/hashicorp.list && \
+  curl \
+    -fs \
+    -o /usr/share/keyrings/hashicorp-archive-keyring.asc \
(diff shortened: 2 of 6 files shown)


anarazel:
Packer changed their license, that's why newer versions aren't in Debian. I'm somewhat hesitant to upgrade to the BSL licensed packer, as this would do.

What exactly is the problem that requires upgrading packer? Perhaps we can work around it some other way.

Eventually I suspect we'll either have to accept the BSL licensed packer, or move to something else :(

nbyavuz:
> What exactly is the problem that requires upgrading packer? Perhaps we can work around it some other way.

Packer can not connect to the FreeBSD 14 instance:

```
packer-builder-googlecompute plugin: [INFO] Attempting SSH connection to ***...
packer-builder-googlecompute plugin: [DEBUG] reconnecting to TCP connection for SSH
packer-builder-googlecompute plugin: [DEBUG] handshaking with SSH
packer-builder-googlecompute plugin: [DEBUG] SSH handshake err: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
packer-builder-googlecompute plugin: [DEBUG] Detected authentication error. Increasing handshake attempts.
packer-builder-googlecompute plugin: [INFO] Attempting SSH connection to ***...
```

anarazel:
> Packer can not connect to the FreeBSD 14 instance:
> ..
> packer-builder-googlecompute plugin: [DEBUG] SSH handshake err: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain

Have you tried to figure out why this is?

@macdice any idea ^  This apparently didn't happen with fbsd 13 and does with 14?

anarazel:
@nbyavuz Looks like packer 1.9.5 was still MPL licensed:
https://github.com/hashicorp/packer/tree/v1.9.5
Perhaps we could check if the problem is already fixed if we just update to that?

nbyavuz:
> > Packer can not connect to the FreeBSD 14 instance:
> > ..
> > packer-builder-googlecompute plugin: [DEBUG] SSH handshake err: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
> 
> Have you tried to figure out why this is?
> 
> @macdice any idea ^ This apparently didn't happen with fbsd 13 and does with 14?

It turns out that was a simple problem. FreeBSD-14 does not accept SSH-RSA key type by default, so adding `temporary_key_pair_type = "ed25519"` fixed the problem.

Closing that PR.

anarazel:
Hm, so where did you add that in https://github.com/anarazel/pg-vm-images/pull/109/files ?

nbyavuz:
> Hm, so where did you add that in https://github.com/anarazel/pg-vm-images/pull/109/files ?

At the end of `source "googlecompute" "freebsd-vanilla" {}` section in the `packer/freebsd.pkr.hcl` file [1].

[1] https://github.com/anarazel/pg-vm-images/blob/455279b57bd2413fdf85a75ebe7ce33f6eed9f93/packer/freebsd.pkr.hcl#L28

PR #114, opened 2025-05-30
windows: Upgrade VS 2019 to VS 2022

nbyavuz:
Per @macdice's request. New APIs are available in the VS 2022.

Also, Windows scripts are moved to scripts/windows/ folder.

Example Windows tasks that uses VS 2022 compiler: https://cirrus-ci.com/build/5132794680049664

-------

Postgres will automatically start using VS 2022 compiler once this PR is merged as image family being same (pg-ci-windows-ci). We will just need to update Windows tasks' names on the upstream Postgres ([Example commit](https://github.com/nbyavuz/postgres/commit/92f779626b36ff331017d4a9b965bd7568e49601)).

This Windows VS 2022 image is already published under pg-ci-images-dev GCP project. I can set unique family name like pg-ci-windows-ci-vs-2022 under pg-ci-images-dev GCP project to play with image easily before merging this PR, what do you think?


anarazel (approved):
But I guess i can live with it...

PR #118, opened 2025-09-19
mingw: Use UCRT toolchain for mingw-w64

nbyavuz:
Switch CC/CXX to x86_64-w64-mingw32ucrt-* to fix build failure from missing _iswctype_l in mingw-w64 v12 headers.

CI Run that MinGW cross compilation passes: https://cirrus-ci.com/task/5499495296794624

### DIFF
--- a/scripts/linux_debian_install_deps.sh
+++ b/scripts/linux_debian_install_deps.sh
@@ -73,8 +73,8 @@ apt-get -y install --no-install-recommends \
   slapd \
   zstd \
   \
-  g++-mingw-w64-x86-64-win32 \
-  gcc-mingw-w64-x86-64-win32 \
+  g++-mingw-w64-ucrt64 \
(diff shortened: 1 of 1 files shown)


anarazel:
I think we should install both in parallel, otherwise there might be a failure of the current CI setup? 
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

nbyavuz:
> I think we should install both in parallel, otherwise there might be a failure of the current CI setup? -- Sent from my Android device with K-9 Mail. Please excuse my brevity.

Yes. I thought that Debian bookworm has these packages too (apparently it does not); so we can merge this PR and commit related patch on the Postgres at the same time but installing both in parallel is better.

anarazel (changes requested):
As discussed, needs both in parallel.

PR #133, opened 2026-01-08
Try to update packer image to bookworm

anarazel:
Can't upgrade to trixie, due to the license change having lead to packer being removed from newer versions.  Will need to solve that at some point :(

This currently fails, possibly due to podman changes.

### DIFF
--- a/docker/linux_debian_packer
+++ b/docker/linux_debian_packer
@@ -1,4 +1,4 @@
-FROM debian:bullseye
+FROM debian:bookworm
 RUN \
   export DEBIAN_FRONTEND=noninteractive && \
   apt-get -y update && \


nbyavuz:
> > This currently fails, possibly due to podman changes.
> > @nbyavuz Could you check to see what this is about?   ISTM that we should at least make the python script report the stderr of the command after failure - because we capute the output and use check=True, we don't even see what the failure is due to.

There are two different errors:

1- Packer was omitting 'project' text from the disk's URL. Please see: https://github.com/hashicorp/packer-plugin-googlecompute/issues/15

2- New version of Podman is more strict about using storage overlays and since we use Podman in the docker container, we can't stack 'overlay' storage containers. To solve that, we need to use 'vfs' as a storage driver or we need to install 'fuse-overlayfs' and set it is a mount_program in the storage.conf. However, there was no default storage.conf in the docker image and I couldn't find a way to create a default conf, so I think it is better to fix this by using '--storage-driver vfs' option while running the Podman.

For the fixing 1, we need to install packer by using its own repository. So, I think it is better to upgrade image to trixie and do that in the trixie image. I created another PR for it, please see #134.

anarazel:
Closing in favor of #134

PR #134, opened 2026-01-08
Try to update packer image to trixie

nbyavuz:
There are three changes:

1- Packer's licence is changed from MPL to Business Source License.
Packer 1.9.5 is the latest version with the MPL licence, so install this
version manually.

2- New version of Podman is more strict about using storage overlays and
since we use Podman in the docker container, we can't stack 'overlay'
storage containers. To solve that, we need to use 'vfs' as a storage
driver or we need to install 'fuse-overlayfs' and set it is a
mount_program in the storage.conf. However, there was no default
storage.conf in the docker image and I couldn't find a way to create a
default conf, so this problem is fixed by using '--storage-driver vfs'
option while running the Podman.

3- The container which builds Windows image started to fail due to OOM.
Its memory is increased from 256Mi to 512Mi.

Postgres CI run: https://cirrus-ci.com/build/6391641252036608

### DIFF
--- a/docker/linux_debian_packer
+++ b/docker/linux_debian_packer
@@ -1,4 +1,10 @@
-FROM debian:bullseye
+FROM debian:trixie
+# Install Packer 1.9.5, it is the latest version with MPL licence
+# Hardcode the current SHA256 checksum of the Packer binary to ensure we are
+# using the correct version of Packer
+ADD --checksum=sha256:6f8272658a6d606583c2b3deaad272233db6e84a6ee651bf17a0d46d8cea4a9c \
+  https://releases.hashicorp.com/packer/1.9.5/packer_1.9.5_linux_amd64.zip \
+  /tmp/packer.zip
 RUN \
   export DEBIAN_FRONTEND=noninteractive && \
   apt-get -y update && \
@@ -6,9 +12,9 @@ RUN \
   apt-get -y --no-install-recommends install \
     ca-certificates \
     curl \
-    packer \
     podman \
     python3 \
+    unzip \
     qemu-system \
     qemu-utils \
     colorized-logs \
@@ -24,4 +30,10 @@ RUN \
   apt-get install --no-install-recommends -y \
     google-cloud-sdk \
     && \
+  unzip /tmp/packer.zip -d /usr/local/bin/ && \
+  rm /tmp/packer.zip && \
+  chmod +x /usr/local/bin/packer && \
+  # Install required Packer plugins to fix 'Bundled plugins used' warnings
+  packer plugins install github.com/hashicorp/googlecompute && \
+  packer plugins install github.com/hashicorp/qemu && \
   apt-get clean && rm -rf /var/lib/apt/lists/*
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -104,10 +104,12 @@ task:
     - env:
         TASK_NAME: freebsd
         PACKERFILE: packer/freebsd.pkr.hcl
+        MEMORY: 256Mi
 
     - env:
         PACKERFILE: packer/linux_debian.pkr.hcl
         SCRIPTS: scripts/linux_debian_*
+        MEMORY: 256Mi
 
       matrix:
         - env:
@@ -129,6 +131,8 @@ task:
         PACKERFILE: packer/windows.pkr.hcl
         SCRIPTS: scripts/windows*
         TASK_NAME: windows-ci
+        # 256Mi is not enough after docker image is updated to trixie
+        MEMORY: 512Mi
 
   env:
     IMAGE_NAME: ${PREFIX}-${TASK_NAME}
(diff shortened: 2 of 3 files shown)


docker/linux_debian_packer
```diff
@@ -24,4 +24,16 @@ RUN \
   apt-get install --no-install-recommends -y \
     google-cloud-sdk \
     && \
+  # Packer is removed from official repositories due to licensing.
+  # Install it from its own repository.
+  echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/hashicorp-archive-keyring.asc] https://apt.releases.hashicorp.com $(grep -oP '(?<=UBUNTU_CODENAME=).*' /etc/os-release || lsb_release -cs) main" > \
```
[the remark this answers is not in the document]
nbyavuz: It worked!

docker/linux_debian_packer
```diff
@@ -24,4 +24,12 @@ RUN \
   apt-get install --no-install-recommends -y \
     google-cloud-sdk \
     && \
+  # Install Packer 1.9.5, it is the latest version with MPL licence
+  curl \
+    -fs \
+    -o /tmp/packer.zip \
+    https://releases.hashicorp.com/packer/1.9.5/packer_1.9.5_linux_amd64.zip && \
```
anarazel: I think it may be good to verify that this matches an expectd sha256sum.

docker/linux_debian_packer
```diff
@@ -24,4 +24,15 @@ RUN \
...
+  unzip /tmp/packer.zip -d /usr/local/bin/ && \
+  rm /tmp/packer.zip && \
+  # Hardcode the current SHA256 checksum of the Packer binary to ensure we are
+  # using the correct version of Packer
+  echo "e290bf0bb9ccda08d905ceddc22992fbd87dad06d6cd91d0d63e50a873b929e7 /usr/local/bin/packer" | sha256sum -c - && \
```
anarazel: This should be the checksum of the packer.zip, not of one of the containing files...

The easiest way to implement that is probably to use https://docs.docker.com/reference/dockerfile/#add---checksum

nbyavuz: Done.

docker/linux_debian_packer
nbyavuz: Done.

anarazel:
Every packer run now has a warning like:
```
[18:22:56.215] packer build \
[18:22:56.215]   -only "*.${TASK_NAME}" \
[18:22:56.215]   -timestamp-ui \
[18:22:56.215]   -force \
[18:22:56.215]   -var "image_date=$DATE" \
[18:22:56.215]   -var "image_name=${IMAGE_NAME}" \
[18:22:56.215]   -var gcp_project=$GCP_PROJECT \
[18:22:56.215]   "$PACKERFILE"
[18:22:56.320] Warning: Bundled plugins used
[18:22:56.320] 
[18:22:56.320] This template relies on the use of plugins bundled into the Packer binary.
[18:22:56.320] The practice of bundling external plugins into Packer will be removed in an
[18:22:56.320] upcoming version.
[18:22:56.320] 
[18:22:56.320] To remove this warning, add the following section to your template:
[18:22:56.320] 
[18:22:56.320] packer {
[18:22:56.320]   required_plugins {
[18:22:56.320]     googlecompute = {
[18:22:56.320]       source  = "github.com/hashicorp/googlecompute"
[18:22:56.320]       version = "~> 1"
[18:22:56.320]     }
[18:22:56.320]   }
[18:22:56.320] }
```

Have you looked at what it takes to fix that? I can see fixing that separately from this PR, but we probably should fix it soon-ish.

nbyavuz:
> Every packer run now has a warning like:
> 
> ```
> [18:22:56.215] packer build \
> [18:22:56.215]   -only "*.${TASK_NAME}" \
> [18:22:56.215]   -timestamp-ui \
> [18:22:56.215]   -force \
> [18:22:56.215]   -var "image_date=$DATE" \
> [18:22:56.215]   -var "image_name=${IMAGE_NAME}" \
> [18:22:56.215]   -var gcp_project=$GCP_PROJECT \
> [18:22:56.215]   "$PACKERFILE"
> [18:22:56.320] Warning: Bundled plugins used
> [18:22:56.320] 
> [18:22:56.320] This template relies on the use of plugins bundled into the Packer binary.
> [18:22:56.320] The practice of bundling external plugins into Packer will be removed in an
> [18:22:56.320] upcoming version.
> [18:22:56.320] 
> [18:22:56.320] To remove this warning, add the following section to your template:
> [18:22:56.320] 
> [18:22:56.320] packer {
> [18:22:56.320]   required_plugins {
> [18:22:56.320]     googlecompute = {
> [18:22:56.320]       source  = "github.com/hashicorp/googlecompute"
> [18:22:56.320]       version = "~> 1"
> [18:22:56.320]     }
> [18:22:56.320]   }
> [18:22:56.320] }
> ```
> 
> Have you looked at what it takes to fix that? I can see fixing that separately from this PR, but we probably should fix it soon-ish.

There are two solutions, first is adding the above section to the packer file and other one is having this script:

```
#!/bin/sh

set -e

PLUGINS="
github.com/hashicorp/googlecompute
github.com/hashicorp/qemu
"

for plugin in $PLUGINS; do
    packer plugins install "$plugin"
done
```

and running this script in each task which uses packer (or running this once while generating the docker image but I didn't try this yet). I actually did that but then removed it since it is not required. If you want I can add this script as a part of this PR or as a separate PR?

anarazel:
I would check whether we can do it as part of the docker image generation, so it doesn't have to be done so often.

Re this PR or separately: Either is fine with me.

anarazel (approved):

anarazel:
The added plugin stuff looks good to me. Please merge if the CI run doesn't show issues.

PR #136, opened 2026-01-08
linux: Mark sid tasks as non-critical

anarazel:
The tasks often fail for reasons outside of our control and breakages don't directly affect upstream postgres CI.


anarazel:
Hah. We were already doing that, just wrongly, for some reason.

@nbyavuz is it possible 07530665401fd4dca8c7f1282c51d2a62800538f never worked?

PR #139, opened 2026-03-31
mingw: pin IPC::Run to NJM/IPC-Run-20250809.0

nbyavuz:
MinGW CI tasks started failing after IPC::Run was updated from NJM/IPC-Run-20250809.0 to TODDR/IPC-Run-20260322.0. Uninstall any existing IPC::Run before installing the last known-good version from NJM to avoid the breakage introduced in the newer release.

See:
https://postgr.es/m/CAN55FZ06xanSbJdHe-CurjX_qNuBWZDEvS1kAk36L38YCtZXnw@mail.gmail.com

Related CI Run: https://cirrus-ci.com/task/4732655020802048

### DIFF
--- a/scripts/windows_install_mingw64.ps1
+++ b/scripts/windows_install_mingw64.ps1
@@ -24,7 +24,14 @@ msys 'pacman -Scc --noconfirm'
 
 # Install perl modules to enable tap tests
 msys 'where perl'
-msys '(echo; echo o conf recommends_policy 0; echo notest install IPC::Run) | cpan'
+echo "Check if IPC::Run is already installed, it shouldn't be at this point"
+msys 'perl -mIPC::Run -e 1 && exit 1 || exit 0'
+# MinGW CI tasks started failing after the package was updated from
+# NJM/IPC-Run-20250809.0 to TODDR/IPC-Run-20260322.0. There is no way to
+# install IPC::Run from an author without specifying the exact version number
+# so install the latest working one. (NJM/IPC-Run-20250809.0.tar.gz). See:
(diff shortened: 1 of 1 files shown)


scripts/windows_install_mingw64.ps1
```diff
@@ -24,7 +24,15 @@ msys 'pacman -Scc --noconfirm'
...
 msys 'where perl'
-msys '(echo; echo o conf recommends_policy 0; echo notest install IPC::Run) | cpan'
+# Remove existing IPC::Run installation if present, then force install from
+# NJM to avoid breakage introduced in TODDR/IPC-Run-20260322.0. MinGW CI
+# tasks started failing after the package was updated from
+# NJM/IPC-Run-20250809.0 to TODDR/IPC-Run-20260322.0. See:
+# - https://postgr.es/m/CAN55FZ06xanSbJdHe-CurjX_qNuBWZDEvS1kAk36L38YCtZXnw%40mail.gmail.com
+# There is no way to install IPC::Run from an author without specifying the
+# exact version number so install the latest working one.
+msys 'if perl -mIPC::Run -e 1 2>/dev/null; then (echo; echo notest uninstall IPC::Run) | cpan; fi'
```
anarazel: Do we need that uninstall?  We control the image, so it really shouldn't be installed already. And if it were, it'd presumably not be installed via cpan?

I'm doubtful it'd work anyway, as on the first run cpan has to be configured (that's what the first "echo;" does in the install invocation).

anarazel: Perhaps just error out if it's already installed?

nbyavuz: Done.

PR #143, opened 2026-08-15
linux: Install a 32bit python in the 32bit image

JelteF:
This is a requirement to get pytest working on our 32bit CI jobs. The 64bit python version cannot load the 32bit libpq library. Installing a 32bit python sadly requires a bit more logic than a simple package install, because the 32bit python package cannot be installed at the same time as the 64bit one. So instead this extracts the 32bit executable from the deb file, and puts it in `/usr/local/bin/python3-i386`. A small helper script to run pytest using this python version is also added there.


scripts/linux_debian_install_deps.sh
```diff
@@ -248,6 +248,35 @@ echo "Installing ${packages[@]}"
...
+    # still useful, because meson's PYTEST option takes a program path and so
+    # cannot spell "python3-i386 -m pytest" itself.
+    cat > /usr/local/bin/pytest-i386 <<'PYTEST_I386'
+#! /bin/sh
+exec /usr/local/bin/python3-i386 -m pytest "$@"
+PYTEST_I386
+    chmod 755 /usr/local/bin/pytest-i386
+fi
+
+
```
anarazel, line 279: Uh, this can't be the right approach. I'm quite opposed to this kind of stuff because it's a maintenance burden going forward.

JelteF: With "this" I assume you mean the whole thing, i.e. installing 32bit python by extracting it from the 

The reason that's needed is because we use a 64bit OS to build and test 32bit software. Python 32bit and 64bit packages in debian/ubuntu are not meant to be installed on the same machine.
I can think of two other options:
1. Don't install 32bit python at all. Use 64bit libpq to test against 32bit postgres server. Losing 32bit libpq coverage seems pretty undesirable.
2. Start creating a fully 32bit image and using that for our 32bit tests. I did that in #144 in case you like that approach better.

PR #144, opened 2026-08-19
linux: Build the 32bit image as a native i386 image

JelteF:
The `linux-meson-32` image was an amd64 image that cross built for i386 with `gcc-multilib` and a hand maintained list of `:i386` -dev packages. That already caused problems because `libossp-uuid-dev:i386` and `systemtap-sdt-dev:i386` conflict with their amd64 counterparts and had to be left out. For the effort to introduce pytest support in our test suite the same thing holds true, a 32bit and 64bit python executable cannot be installed together.

To solve that problem this starts basing the image on `i386/debian` instead. That way we actually get a whole 32bit OS instead of one that's half 32bit.

Related to #143

### DIFF
--- a/docker/linux_debian_ci
+++ b/docker/linux_debian_ci
@@ -59,15 +59,50 @@ RUN \
 FROM base AS sanity-check
 
 
-# Image for the linux-meson-32 job
-FROM normal AS linux-meson-32
+# Image for the linux-meson-32 job.
+#
+# This is a native i386 image, rather than an amd64 image cross building with
+# gcc-multilib and a hand maintained list of :i386 -dev packages. The package
+# list is then simply the same one as for the other jobs, and everything that
+# runs during the tests (python/pytest in particular) is 32bit as a matter of
+# course, rather than something we have to arrange for. It also means the job
+# tests a real i386 build, which is what the buildfarm members we care about
+# are doing.
+#
+# The base/normal steps are repeated here rather than shared with the amd64
+# stages above, because a stage's FROM cannot be varied per --target within a
+# single build.
+FROM i386/debian:$DEBIAN_RELEASE AS base-32
+
+LABEL \
+  org.opencontainers.image.source="https://github.com/anarazel/pg-vm-images" \
+  org.opencontainers.image.description="image for running postgres CI"
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV IMAGE_NAME=$IMAGE_NAME
+ENV IMAGE_DATE=$DATE
+ENV IMAGE_IDENTITY=$IMAGE_NAME-$DATE
+
+COPY /scripts/linux_debian_install_deps.sh \
+  /scripts/update_ipc_run.sh \
+  /scripts/
+
+FROM base-32 AS linux-meson-32
 
 RUN \
   --mount=type=cache,target=/var/cache/apt,sharing=locked \
   --mount=type=cache,target=/var/lib/apt,sharing=locked \
   \
+  rm -f /etc/apt/apt.conf.d/docker-clean && \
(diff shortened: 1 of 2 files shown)


docker/linux_debian_ci
```diff
@@ -59,15 +59,50 @@ RUN \
...
+# list is then simply the same one as for the other jobs, and everything that
+# runs during the tests (python/pytest in particular) is 32bit as a matter of
+# course, rather than something we have to arrange for. It also means the job
+# tests a real i386 build, which is what the buildfarm members we care about
+# are doing.
+#
+# The base/normal steps are repeated here rather than shared with the amd64
+# stages above, because a stage's FROM cannot be varied per --target within a
+# single build.
+FROM i386/debian:$DEBIAN_RELEASE AS base-32
```
nbyavuz: This fails with:

```
  [4/5] STEP 1/7: FROM i386/debian:trixie AS base-32
  Resolving "i386/debian" using unqualified-search registries (/etc/containers/registries.conf)
  Trying to pull docker.io/i386/debian:trixie...
  Trying to pull quay.io/i386/debian:trixie...
  Error: creating build container: 2 errors occurred while pulling:
   * unable to copy from source docker://i386/debian:trixie: choosing an image from manifest list docker://i386/debian:trixie: no image found in image index for architecture "amd64", variant "", OS "linux"
   * unable to copy from source docker://quay.io/i386/debian:trixie: initializing source docker://quay.io/i386/debian:trixie: reading manifest trixie in quay.io/i386/debian: unauthorized: access to the requested resource is not authorized
```

`--platform=linux/386` needs to be added:

`FROM --platform=linux/386 i386/debian:$DEBIAN_RELEASE AS base-32`

nbyavuz: Also,

`if: ${{ matrix.variant == 'linux-meson-32' }}`

line in the `container.yml` file needs to be updated with:

`if: ${{ matrix.variant == 'linux-meson-64' }}`

nbyavuz, line 75: I think this is a better solution compared to #143 but there are some problems.

Image building fails with:

```
  [4/5] STEP 1/7: FROM i386/debian:trixie AS base-32
  Resolving "i386/debian" using unqualified-search registries (/etc/containers/registries.conf)
  Trying to pull docker.io/i386/debian:trixie...
  Trying to pull quay.io/i386/debian:trixie...
  Error: creating build container: 2 errors occurred while pulling:
   * unable to copy from source docker://i386/debian:trixie: choosing an image from manifest list docker://i386/debian:trixie: no image found in image index for architecture "amd64", variant "", OS "linux"
   * unable to copy from source docker://quay.io/i386/debian:trixie: initializing source docker://quay.io/i386/debian:trixie: reading manifest trixie in quay.io/i386/debian: unauthorized: access to the requested resource is not authorized
```

`--platform=linux/386` needs to be added:

`FROM --platform=linux/386 i386/debian:$DEBIAN_RELEASE AS base-32`

----

Then, image build successfully finishes but there are other problems when this image is used:

When Github tries to checkout, it fails with:

```
Run actions/checkout@v6
/usr/bin/docker exec  81cb150397cf626447f9a98084b7c195b98ca3ea8d8ca144cf98d7441de40947 sh -c "cat /etc/*release | grep ^ID"
exec /__e/node24/bin/node: no such file or directory
```
because GitHub Actions uses 64-bit Node.js binaries. I did a bit research and found that people install 64-bit node environment to their images as solution but that is similar to the problem we are trying to solve (I tried this solution and it works [1]).

----

Also,

`if: ${{ matrix.variant == 'linux-meson-32' }}`

line in the `container.yml` file needs to be updated with:

`if: ${{ matrix.variant == 'linux-meson-64' }}`

----

[1] https://github.com/nbyavuz/pg-vm-images/actions/runs/32459538054
[1] https://github.com/nbyavuz/postgres/actions/runs/32460233796

